PR #3, opened 2017-07-17
Docstrings for mtf-generator and common libs

alexxa:
This PR updates docstrings for mtf-generator and common libs.


jscotka:
LGTM

PR #5, opened 2017-07-17
fex fixes solves issues with chnaged names of current function

jscotka:
it fixes several issues caused by rewriting to  better function names

### DIFF
--- a/examples/multios_testing/sanity1.py
+++ b/examples/multios_testing/sanity1.py
@@ -39,11 +39,11 @@ class SanityCheck1(module_framework.AvocadoTest):
         self.run(" nohup nc -l %d > %s 2>&1 &" % (PORT, TFILE), shell=True, ignore_bg_processes=True)
         time.sleep(2)
         a = self.runHost("cat /etc/redhat-release").stdout.strip()
-        self.runHost("cat /etc/redhat-release | nc localhost %d" % PORT, shell=True)
+        self.runHost("Im Host | nc localhost %d" % PORT, shell=True)
         time.sleep(2)
         b = self.run("cat %s" %TFILE).stdout.strip()
         self.assertEqual(a,b)
-        self.assertIn("26", b)
+        self.assertIn("Im Host", b)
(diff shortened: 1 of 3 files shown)


examples/multios_testing/sanity1.py
```diff
@@ -39,11 +39,11 @@ def testServerInside(self):
         self.run(" nohup nc -l %d > %s 2>&1 &" % (PORT, TFILE), shell=True, ignore_bg_processes=True)
         time.sleep(2)
         a = self.runHost("cat /etc/redhat-release").stdout.strip()
-        self.runHost("cat /etc/redhat-release | nc localhost %d" % PORT, shell=True)
+        self.runHost("Im Host | nc localhost %d" % PORT, shell=True)
```
phracek: Can you please explain it, what do you want to test?
I don't understand at all.

moduleframework/module_framework.py
```diff
@@ -104,6 +104,18 @@ def runHost(self, command="ls /", **kwargs):
                 trans_dict)
         return process.run("%s" % formattedcommand, **kwargs)
 
+    def PackagesInstalled(self,packagelist, func="runHost"):
```
phracek: Why empty function?

examples/multios_testing/sanity1.py
jscotka: I'm testing difference between test host from host and guest from host. And in previous version there were dependency on host fedora version -> this leads to errors because version increased from 25->26 so this text is better.
This host from host test does not so big sense as guest from host, but it shows ability, that just run command is replaced by runHost and it is testing env vice-versa

moduleframework/module_framework.py
jscotka: Because it should speedup testing, but this caused opposite effect, so that I've let function there, for further work, but I've let it empty.

jscotka:
It is too old. Closing. many issues solved by another PRs.

PR #17, opened 2017-07-24
Use dnf instead of microdnf

phracek:

I totally forgot to mention reason.
In Dockerfile file we test whether it uses microdnf and not dnf. Unfortunately, it was changed.
Now we should check whether Dockerfile contains dnf and not microdnf. dnf turn on modularity in Dockerfiles.


ignatenkobrain:
somehow I miss description of commit message to understand why is it needed.

phracek:
This will be replaced the other Pull Request. Let's check clean all and not packagers itself.

PR #24, opened 2017-07-26
Remove Internal Jenkins instance

phracek:
This fixes only document.
Instead of Internal Jenkins document use Arbitrary Jenkins.


TomasTomecek:
lgtm

PR #32, opened 2017-08-07
obsolete modularity-testing-framework-0.5.18-1, too

nphilipp:


meta-test-family.spec
```diff
@@ -23,7 +23,7 @@ Requires:       python2-dockerfile-parse
 Requires:       python2-pdc-client
 Requires:       python2-modulemd
 Provides:       modularity-testing-framework = %{version}-%{release}
-Obsoletes:      modularity-testing-framework < 0.5.18-1
+Obsoletes:      modularity-testing-framework <= 0.5.18-1
```
ignatenkobrain: actually, it should be `< 0.5.18-2`, because `0.5.18-1 < 0.5.18-1.fc26`

nphilipp: Of course :).

PR #42, opened 2017-08-10
Fixes #30 Remove shebags

### LINKED ISSUE #30: Update SPEC file according new MTF name
#28 updates documentation including SPEC file.
Old name SPEC name was modularity-testing-framework.spec with name modularity-testing-framework

How to update SPEC file according to new name.
There are two options how to handle it:

1. keep the same name and add to spec file http://pkgs.fedoraproject.org/cgit/rpms/coreutils.git/tree/coreutils.spec?h=f26#n130
    In our case: Provides: meta-test-family = %{version} - %{release}
1. Mark modularity-testing-framework as obsolete and introduce the new name. I don't know in my mind, If the new package request will be needed. Including new Fedora dist-git repo.

Conversation from #28 
@phracek , I updated copr repo name in docs... As per spec, I would go with the option number two. And I also would change all occurrences of moduleframework to mtf, agree? If so, I will proceed.
@alexxa Let's discuss it together with other folks. I will create a new issue, so we can merge documentation.

What do you thing folks? @ignatenkobrain @TomasTomecek

phracek:

this Pull Request removes shebang from Python files, which are not needed.


phracek:
@jscotka @TomasTomecek Please have a look on it.

meta-test-family.spec
```diff
@@ -22,11 +22,26 @@ Requires:       docker
 Requires:       python2-dockerfile-parse
```
ignatenkobrain: this should be moved to python2- subpackage, no?

phracek: Right.

moduleframework/mtf_generator.py
```diff
@@ -99,6 +99,3 @@ def main():
     configout = open('generated.py', 'w')
     configout.write(config.output)
     configout.close()
-
-if __name__ == '__main__':
```
TomasTomecek: Is this intended?

moduleframework/timeoutlib.py
```diff
@@ -148,120 +147,3 @@ def __wrap(*args, **kwargs):
                 delay = self.handle_failure(start_time, "")
 
         return __wrap
-
-
-if __name__ == '__main__':
```
TomasTomecek: huh, just like that?

meta-test-family.spec
```diff
@@ -19,14 +19,28 @@ Requires:       python2-avocado
 Requires:       python2-avocado-plugins-output-html
 Requires:       python-netifaces
```
TomasTomecek: I think the other runtime requires should also be handled by the python2- subpkage. Haven't check the executables though.

moduleframework/mtf_generator.py
phracek: You are right. I did more changes then remove shebangs

examples/testing-module/moduleframework-cmd
```diff
@@ -1 +1 @@
-../../moduleframework/bashhelper.py
\ No newline at end of file
+../../tools/bashhelper.py
```
alexxa: This line requires docs update: api/bashhelper.rst:4:.. automodule:: moduleframework.bashhelper

PR #44, opened 2017-08-10
Split prepare and cleanup for environment

jscotka:
it is Work In progress,
It allows to split environment prepare and cleanup from tests, so that. it is done just ones time, on demand.
please do not merge it now.

### DIFF
--- a/docs/example-config-minimal.yaml
+++ b/docs/example-config-minimal.yaml
@@ -5,7 +5,6 @@ modulemd-url: https://src.fedoraproject.org/modules/memcached/raw/master/f/memca
 packages:
     rpms:
         - bash
-        - rpm
 module:
     docker:
         container: docker.io/phracek/memcached
--- a/docs/example-config-minimal.yaml
+++ b/docs/example-config-minimal.yaml
@@ -7,7 +7,7 @@ packages:
         - bash
 module:
     docker:
-        container: docker.io/phracek/memcached
+        container: docker.io/modularitycontainers/memcached
     rpm:
         repo: https://kojipkgs.fedoraproject.org/compose/latest-Fedora-Modular-26/compose/Server/x86_64/os/
 
--- a/setup.py
+++ b/setup.py
@@ -78,6 +78,8 @@ setup(
             'moduleframework-cmd = moduleframework.bashhelper:main',
             'modulelint = moduleframework.modulelint:main',
             'mtf-generator = moduleframework.mtf_generator:main',
+            'mtf-env-set = moduleframework.mtf_environment:mtfenvset',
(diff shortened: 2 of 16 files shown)


phracek:
@jscotka Would it be possible to write steps how to use it in this PR?

jscotka:
Hi @phracek @alexxa 
usage is as we discussed:
```
MODULE=docker mtf-env-set
MODULE=docker avocado run test1.py
MODULE=docker mtf-env-clean
```

Now it is ready for review and merging

docs/example-config-minimal.yaml
```diff
@@ -5,7 +5,6 @@ modulemd-url: https://src.fedoraproject.org/modules/memcached/raw/master/f/memca
 packages:
     rpms:
         - bash
-        - rpm
 module:
     docker:
         container: docker.io/phracek/memcached
```
phracek: I would prefer to use docker.io/modularitycontainers/memcached

setup.py
```diff
@@ -78,6 +78,8 @@ def get_dir(system_path=None, virtual_path=None):
             'moduleframework-cmd = moduleframework.bashhelper:main',
             'modulelint = moduleframework.modulelint:main',
             'mtf-generator = moduleframework.mtf_generator:main',
+            'mtf-env-set = moduleframework.mtf_environment:mtfenvset',
```
phracek: Could it be possle to use instead of avocado run *.py something like just mtf which will do this?

phracek:
Thanks for the update.
Can we change 
~~~
MODULE=docker avocado run test1.py
~~~
to something like this:
~~~
MODULE=docker mtf
~~~

mtf will just run avocado run *.py tests. Does it make sense?
Otherwise I am fine with this PR.

setup.py
jscotka: It is possible. But I'm not sure if it is generally good idea.
Because:
 * run all *.py files (there may be also not tests, some common python files. It leads to errors when scheduling or when you have there some tests not ready to schedule.
 * it is more complicate found which tests will be scheduled
 * I prefer to have explicit set of tests in production like in makefile TESTS=sanity1.py sanity2.py smoke.py (instead of *.py) it is also more compatible with ansible standard test roles. They want to have explicit test set.
 * also we support also behave, or frameworks what you choose. so that. mtf -> can be bash alias like: alias mtf="avocado run" but it is just trivial shortcut

So yep, I can imagine that mtf could be alias to "avocado run",  but not to avocado run *.py, because it can leads to bad behaviour.

jscotka:
this change could be part of another PR, what will add command mtf, what could be alias to avocado run and @alexxa suggested to add also mtf --version command

PR #48, opened 2017-08-14
example of tests for collections

jscotka:
Test for collections, example how to use it directly, without any modification

### DIFF
--- a/examples/rhscl_maria/config.yaml
+++ b/examples/rhscl_maria/config.yaml
@@ -1,0 +1,18 @@
+document: modularity-testing
+version: 1
+name: mariadb
+modulemd-url: https://src.fedoraproject.org/modules/base-runtime/raw/master/f/base-runtime.yaml
+packages:
+    rpms:
+        - bash
+testdependencies:
+    rpms:
+        - nc
+        - community-mysql
+default_module: docker
+module:
+    docker:
+        start: docker run -p 3306:3306 -e MYSQL_ROOT_PASSWORD={PASSWORD}
+        container: docker.io/mariadb
+    rpm:
+        source: http://ftp.fi.muni.cz/pub/linux/fedora/linux/releases/26/Everything/x86_64/os/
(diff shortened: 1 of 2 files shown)


alexxa:
@jscotka, before reviewing it all
1) Is there any docs should be added to follow these examples? 
2) Let's add docstrings to our test examples, at least some basic ones, see example: https://github.com/RedHatQE/rhui3-automation/blob/master/tests/rhui3_tests/test_entitlements.py
If test expect some value, that should be in a docstring.

examples/rhscl_maria/config.yaml
```diff
@@ -0,0 +1,18 @@
+document: modularity-testing
+version: 1
+name: mariadb
+modulemd-url: https://src.fedoraproject.org/modules/base-runtime/raw/master/f/base-runtime.yaml
```
TomasTomecek: base-runtime.yaml and mariadb, that doesn't seem to match

jscotka:
Hi @alexxa 
Thanks for suggestions. This is just example, but yes, it is better to describe what tests are doing for wider audience. There aren't expected any special values.

examples/rhscl_maria/example1.py
```diff
@@ -0,0 +1,119 @@
...
+from moduleframework import common
+from avocado.utils import process
+from avocado import Test
+import time
+
+WAIT_TIME=10
+
+class OneMachine(module_framework.AvocadoTest):
+    """
+    Basic, the most straightforward example how to use it with collections,
```
TomasTomecek: are these tests meant to be run on rpms or containers?

examples/rhscl_maria/config.yaml
jscotka: Hi @TomasTomecek it is more less just stub, because I wanted to show to hhorak, how to use MTF for testing collections. And now I expect that there will be some modulemd file (although empty) so that I've choosed some generic, and baseruntime is the most minimal one. This example allows to test collection without any changes of MTF

examples/rhscl_maria/example1.py
jscotka: It is just for containers. I should use module_framework.ContainerAvocadoTest instead of generic one. I'll change it.

alexxa:
@jscotka If @TomasTomecek agrees, I think it can be merged.

PR #51, opened 2017-08-21
check if travis works as expected 

jscotka:
dnf inside module, not apt, it is just env packager


jscotka:
Yay,  it works :-)

jscotka:
works, closing this PR without merge.

PR #57, opened 2017-08-23
example's test needs update

psklenar:
test in example has an issue:
self.cancel() has to be used otherwise it has traceback


alexxa:
@psklenar Thanks for a commit, just next time please add more info, e.g. a traceback would be appreciated here.

PR #63, opened 2017-08-24
Fixing packaging issues

### LINKED ISSUE #22: Notes on packaging
Here are some notes/questions that I think might be useful to discuss/think through:

* What is the reason to put modulelint it data files (/usr/share)? AFAICS it's just another Python module, so it should go alongside `moduleframework` or live inside it, IMO.
* I'd recommend creating a `requirements.txt` file, which is a de-facto standard in Python community to track dependencies.
* I don't understand following line in the spec: `chmod a+x %{buildroot}%{python_sitelib}/%{framework_name}/{module_framework,mtf_generator,bashhelper,setup}.py` - The `python_sitelib` shouldn't contain executable files. Either these files shouldn't be executable at all or they should have their functionality available through an endpoint.

phracek:
The Pull Request removes
- `__main__` functions from python files which are installed under /usr/lib/python2.7/site-package/*
- It includes `requirements.txt`


phracek:
This Pull Request fixes issue #22

bkabrda:
So some notes:
* I don't think you need to remove the `__main__` stuff, it can be quite useful for local development/testing, but I'll leave that up to you.
* The `requirements.txt` don't match what's currently in `setup.py` `install_requires`. What I'd recommend:
  * Fix `requirements.txt` to contain what's in `install_requires` right now
  * Replace `install_requires` in `setup.py` to be `open('requirements.txt').read().splitlines()`
  * Add `requirements.txt` to `MANIFEST.in`
* AFAICS this only fixes one of the 3 issues mentioned in #22, so I wouldn't say it fixes that issue.

phracek:
Well. `meta-test-family.spec` does not contain any `chmod a+x` see here. https://github.com/fedora-modularity/meta-test-family/blob/devel/meta-test-family.spec#L40

Master branch still contains the bug, but we are waiting for merge from devel into master. https://github.com/fedora-modularity/meta-test-family/blob/master/meta-test-family.spec#L42

Python files mentioned in /usr/share/moduleframework/tools are used for alone testing like here:
https://github.com/fedora-modularity/meta-test-family/blob/devel/examples/testing-module/Makefile#L2
https://src.fedoraproject.org/modules/mongodb/blob/master/f/tests/Makefile#_1

I think moving them into `/usr/lib/python2.7/site-packages/moduleframework/tools/modulelint/modulelint.py` is a bit nightmare. 
The best propose would be to have a scripts like `/usr/bin/mtf-test-modulelint` but I don't know whether it will work. This needs to investigate a bit.

bkabrda:
Ok, the spec has been moved and heavily modified since #22 was opened, so it looks good now.

As for the `/usr/share` files - I think this is exactly the UX problem of MTF - it lacks a single entrypoint to which you could pass some arguments and it substitutes this with Makefiles that everyone has to duplicate. If you had a single `mtf` binary, you could add an argument like `--run-default-tests` and noone would have to care where these files actually live. Plus, if they lived in `site-packages`, they'd actually be importable by MTF users without having to modify `PYTHONPATH`.

One more minor nit - I don't think that "behave" is actually a hard dependency of MTF and my guess is that most MTF users won't use it, so my proposal would be to remove that from requirements.txt, but I don't think it's a huge issue.

bkabrda:
LGTM now, I see no further issues.

PR #20, opened 2017-07-25
Tests whether dnf contains --nodocs and clean all

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

This Pull Request tests whether dnf or yum contains --nodocs and clean all commands in RUN section.
testing of microdnf was removed. It does not make sense at all. Container developer can use either microdnf, dnf or even yum in CentOS word.

PR #77, opened 2017-08-30
Revert "Tests whether dnf contains --nodocs and clean all"

ignatenkobrain:
Reverts fedora-modularity/meta-test-family#20

It is not complete and doesn't do right job (see original PR for comments).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -254,7 +254,7 @@ man_pages = [
 # (source start file, target name, title, author,
 #  dir menu entry, description, category)
 texinfo_documents = [
-  (master_doc, 'MetaTestfamily', u'Meta Test family Documentation',
+  (master_doc, 'MetaTestFamily', u'MetaTestFamily Documentation',
    u'Petr Hracek@*Jan Scotka', 'MetaTestFamily', 'One line description of project.',
    'Miscellaneous'),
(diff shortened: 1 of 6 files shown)


phracek:
You are saying in PR what is wrong. That's fine, but related to getting yum/dnf I don't see right way. Only this is not correct. What is correct approach?

The same for `dnf clean all` or `yum clean all`

Regarding documents. I am checking whether package is really installed in container and If so, then I check whether it really does not contain document.

If this is wrong, let's fix it and not revert PR.

ignatenkobrain:
@phracek why did we merge PR in first place even I requested changes there?

jscotka:
Otherwise @ignatenkobrain @phracek . 
please use {GUESTPACKAGER} inside the command, every run command is formated by using https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/common.py#L347 trans_dict from https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/common.py#L58
what sets what packager to use, yum, dnf, apt, etc,
Implementation is not perfect, but at least it is good to have it on one place, how to decide what command to use.

phracek:
accidentally I remove the code. I will file a new PR.

PR #74, opened 2017-08-29
Module type split

jscotka:
Module types are now independent, and be able to run alone.
removed mandatory modulemd url -> no module testing
I expect to have some ideas, what to improve, if any.
This PR contains big changes of MTF.
Contains performance improvements and changes according to module container ideas

PR #78, opened 2017-08-31
Merge devel and master

alexxa:


alexxa:
@phracek .... we need to cooperate more, @jscotka wanted to merge it after #74  =((

PR #80, opened 2017-09-05
examples for pytest, unittest, nosetest - simple examples

jscotka:
For people who are interested to use alternative framework than avocado, there are  some examples for

- nosetest
- py.test
- unittest

just simple one. when someone ask, there could be added some derived classes based on these frameworks, to avoid explicit calls for backends function especially in setUp or tearDown.
Especially @psklenar were interested how to use it.

you can run it via `Makefile`
there are targets for these framworks `check-*` sections in `Makefile`

### DIFF
--- a/examples/alternative_frameworks/Makefile
+++ b/examples/alternative_frameworks/Makefile
@@ -1,0 +1,16 @@
+export MTF_REMOTE_REPOS=yes
+export DEBUG=yes
+export MODULE=docker
+
+prepare-docker:
+	 dnf -y install pytest python2-case
+	 mtf-env-set
+
+check-nosetest: prepare-docker
+	nosetests
+
+check-unittest: prepare-docker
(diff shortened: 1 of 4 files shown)


alexxa:
@jscotka any docs should be added maybe?

jscotka:
Hi, @alexxa. what do you suggest?
These examples are  as simple as possible. (not testing anything useful, just check that ls / command inside container return sbin directory as output)
We can add there some docs and inline code comments to explain why each line is important.

phracek:
@alexxa @jscotka I would like to create a new section in docu called like `examples`, where could be examples how to use MTF.

Either directly in http://meta-test-family.readthedocs.io/en/latest/ or may be in Users Guide http://meta-test-family.readthedocs.io/en/latest/user_guide/index.html

PR #94, opened 2017-09-11
Remove BaseRuntime test

### LINKED ISSUE #73: baseruntime check
Why are all images required to be based on `baseruntime/baseruntime`? What if I don't want to use baseruntime at all or my image is based on image that is based on baseruntime (i.e. I do use baseruntime transitively)?

kosciCZ:
Fixes #73


phracek:
LGTM. Thanks

PR #97, opened 2017-09-12
Nodocs clean all check

phracek:
I have removed some tests in Docker by PR https://github.com/fedora-modularity/meta-test-family/pull/63.

This new PR brings two thinks.
* introduce back tests for nodocs and clean_all in Dockerfile
* fix python function names in modulelint.py


moduleframework/tools/modulelint.py
```diff
@@ -43,57 +43,75 @@ def setUp(self):
...
+
+    """
+
+    def test_docker_nodocs(self):
+        self.start()
+        installed_pkgs = self.run("rpm -qa --qf '%{{NAME}}\n'", verbose=False).stdout
+        # This returns a list of packages defined in config.yaml for testing
+        # e.g. ["bash", "rpm", "memcached"] in case of memcached
+        pkgs = self.backend.getPackageList()
+        list_pkg = [pkg for pkg in installed_pkgs.split('\n') if pkg in pkgs]
```
TomasTomecek: would be much easier to compare them as sets: `set(installed_pkgs).intersection(set(defined_packages))`

moduleframework/tools/modulelint.py
```diff
@@ -43,57 +43,75 @@ def setUp(self):
...
-            self.assertTrue("%s" % exp in label_list[label_io_openshift])
+class DockerfileLinterInContainer(container_avocado_test.ContainerAvocadoTest):
+    """
+    :avocado: enable
+
+    """
+
+    def test_docker_nodocs(self):
+        self.start()
+        installed_pkgs = self.run("rpm -qa --qf '%{{NAME}}\n'", verbose=False).stdout
```
ignatenkobrain: just replace by `test -e $(rpm -qad)`

moduleframework/tools/modulelint.py
```diff
@@ -43,57 +43,75 @@ def setUp(self):
...
+        distro = self.run("cat /etc/os-release").stdout
 
-    def testIoOpenShiftTagsExists(self):
-        label_list = self.dp.get_docker_labels()
-        self.assertTrue("io.openshift.tags" in label_list)
+        if 'NAME=Fedora' in distro:
+            pkg_mgr = "dnf"
+        # Look, whether we have solv files in /var/cache/<pkg_mgr>/*.solv
+        # dnf|yum clean all deletes the file *.solv
+        ret = self.run("ls /var/cache/%s/*.solv" % pkg_mgr, ignore_status=True)
```
ignatenkobrain: as I said previously, this check is totally wrong. yum doesn't have solv files for example. also this ties tests to fedora.

TomasTomecek: doesn't work on RHEL, there are no `.solv` files in there:

```
[root@b06253698a54 /]# yum makecache
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
extras               | 3.0 kB  00:00:00
rhel-7-server-rpms   | 4.1 kB  00:00:00
Metadata Cache Created

[root@b06253698a54 /]# cd /var/cache/yum

[root@b06253698a54 yum]# ll
total 0
drwxr-xr-x. 3 root root 21 Oct  4 11:34 x86_64

[root@b06253698a54 yum]# ll x86_64/7Server/
total 16
drwxr-xr-x. 4 root root 4096 Oct  9 09:17 extras
drwxr-xr-x. 4 root root 4096 Oct  9 09:17 rhel-7-server-rpms
-rw-r--r--. 1 root root   46 Oct  9 09:17 timedhosts

[root@b06253698a54 yum]# du -sh .
106M    .

[root@b06253698a54 yum]# yum clean all
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
Cleaning repos: extras rhel-7-server-rpms
Cleaning up everything
Maybe you want: rm -rf /var/cache/yum, to also free up space taken by orphaned data from disabled or removed repos

[root@b06253698a54 yum]# du -sh .
8.0K    .
```

I would probably check via `du` whether `/var/cache/yum` exceeds certain size, something like `128K`.

moduleframework/tools/modulelint.py
phracek: This is questionable whether we would like to test only packages delivered by module or the whole system including Platform.
For some cases this would be a tricky and we need to have a whitelist.
Personally, I don't want to have this test personally failing because of e.g Platform will contain some documents which will be needed. What do you think? Does it make sense?

ignatenkobrain: no, there should be no whitelist. if there are docs installed, this container is not following guidelines. whether docs are coming from different modules doesn't matter because we are testing single unit.

moduleframework/tools/modulelint.py
phracek: Fixed. Thanks

moduleframework/tools/modulelint.py
phracek: This is fixed by some deterministic size, like proposing @TomasTomecek

moduleframework/tools/modulelint.py
TomasTomecek: I think this test is a good candidate for a warning and not an error. To support Petr, here is a brief analysis:

```
$ docker run -ti registry.fedoraproject.org/fedora-minimal:26 bash
bash-4.4# rpm -qad | wc -l
544

$ docker run -ti docker.io/jamesantill/boltron-27 bash
Boltron-27-2ddd828be4c9-0# rpm -qad | wc -l
5529

$ docker run -ti registry.access.redhat.com/rhel7 bash
[root@025a6e693f15 /]# rpm -qad | wc -l
1951
```

Igor, I agree with you that it would be great to test complete images and have
them valid. But as you can see, all our base images contain tons of
documentation. Therefore maybe we could have 2 use cases covered here (by 2
different tests):
- `rpm -qad` which would produce a warning in case of failure
- `rpm -qd $PACKAGES_INSTALLED_IN_THIS_IMAGE` which could produce an error

WDYT?

ignatenkobrain: I think that you are doing wrong thing. RPM will always show you docs, but you don't test files for existence.

TomasTomecek: wow, haven't realized that, good point Igor! new findings then:

```
$ docker run -ti registry.fedoraproject.org/fedora-minimal:26 bash
bash-4.4# for f in $(rpm -qad); do test -e $f && echo $f; done
/usr/share/info/dir

$ docker run -ti docker.io/jamesantill/boltron-27 bash                                                                                                         1 ↵
Boltron-27-d12f22f5458d-0# for f in $(rpm -qad); do test -e $f && echo $f; done | wc -l
4037

$ docker run -ti registry.access.redhat.com/rhel7 bash
[root@23e1d08c9da0 /]# for f in $(rpm -qad); do test -e $f && echo $f; done
/usr/share/info/dir

[root@23e1d08c9da0 /]# rpm -qf /usr/share/info/dir
info-6.5-1.fc28.x86_64

[root@23e1d08c9da0 /]# cat /usr/share/info/dir
This is the file .../info/dir, which contains the
topmost node of the Info hierarchy, called (dir)Top.
The first time you invoke Info you start off looking at this node.

File: dir,      Node: Top       This is the top of the INFO tree

  This (the Directory node) gives a menu of major topics.
  Typing "q" exits, "?" lists all Info commands, "d" returns here,
  "h" gives a primer for first-timers,
  "mEmacs<Return>" visits the Emacs manual, etc.

  In Emacs, you can click mouse button 2 on a menu item or cross reference
  to select it.

* Menu:
```

who will fix `info` package?

ignatenkobrain: This is probably bug in RPM rather than in `texinfo` package...

```
%config(noreplace) %verify(not md5 size mtime) %{_infodir}/dir
```

I don't see why it would be marked as %doc. Can you report bug on https://github.com/rpm-software-management/rpm ?

moduleframework/tools/modulelint.py
```diff
@@ -71,25 +71,96 @@ def test_run_or_usage_label_exists(self):
...
+        self.assertTrue(True)
+
+    def test_installed_docs(self):
+        """
+        This test checks whether no docs are installed by RUN dnf command
+        :return: FAILED in case we found some docs
+                 PASS in case there is no doc file found
+        """
+        self.start()
+        installed_pkgs = self.run("rpm -qa --qf '%{{NAME}}\n'", verbose=False).stdout
```
ignatenkobrain: this is wrong, you should not check %{name}. Check the full NEVRA because this is only one unique things

moduleframework/tools/modulelint.py
```diff
@@ -71,25 +71,96 @@ def test_run_or_usage_label_exists(self):
...
+        test_failed = False
+        for pkg in list_pkg:
+            pkg_doc = self.run("rpm -qd %s" % pkg, verbose=False).stdout
+            if self._file_to_check(pkg_doc):
+                test_failed = True
+        self.assertFalse(test_failed)
+
+    def test_docker_clean_all(self):
+        """
+        This test checks if size of /var/cache/<pkg_manager> is bigger then
```
ignatenkobrain: this function is still wrong in my opinion

phracek: What do you prefer? I would like to have something which will work for this time.
Parsing image layers one by one will follow soon.

@ignatenkobrain What is your preference? Please paste code.

moduleframework/tools/modulelint.py
phracek: Yeah, it is a stupid typo. I will fix it with %{NAME} based on rpm man page.

phracek: Unfortunately test takes a pretty much time. Around 150s which is not good. Based on Travis tests.

phracek:
I would like to thank you for your reviews and time which you spend with them.

moduleframework/tools/modulelint.py
phracek: See reason of double brackets:
https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/common.py#L133
https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/common.py#L384

moduleframework/tools/modulelint.py
phracek: @ignatenkobrain I am still waiting for your proposal. I will merge it on Friday, nevertheless.

ignatenkobrain: I told this many times. Do `clean all` unconditionally when building images.

TomasTomecek: @ignatenkobrain the response from the team who is working on the build system was negative - they advised not to do such magic, hence we need to figure out the linter

TomasTomecek:
needs rebase

moduleframework/dockerlinter.py
```diff
@@ -228,6 +228,19 @@ def check_chained_run_rest_commands(self):
             return False
         return True
 
+    def check_clean_all_command(self):
+        """
+        Function checks if Dockerfile contains dnf/yum clean all in RUN instruction,
```
bkabrda: Suggestion for better docstring:

```
This function checks whether every RUN instruction containing a dnf/yum operation ends with a "dnf/yum clean all".

:return: True if every dnf/yum instruction contains a cleanup step
         False otherwise
```
```

File: moduleframework/tools/dockerlint.py [hunk 7]
```diff
@@ -97,6 +99,81 @@ def test_helpmd_is_present(self):
...
+        for pkg in list_pkg:
+            pkg_doc = self.run("rpm -qd %s" % pkg, verbose=False).stdout
+            if self._file_to_check(pkg_doc):
+                test_failed = True
+        self.assertFalse(test_failed)
+
+    def test_docker_clean_all(self):
+        """
+        This test checks if size of /var/cache/<pkg_manager> is bigger then
+        150000 taken by command du -hsb /var/cache/<pkg_manager>
```
bkabrda: Why 150000? I just tried creating an image where I installed a package from random copr repo and even though I did `dnf clean all`, the directory had `6537257`, so it clearly wouldn't pass this check even though it's correct.

moduleframework/tools/dockerlint.py
```diff
@@ -97,6 +99,81 @@ def test_helpmd_is_present(self):
...
+            exit_status = self.run("test -e %s" % doc, ignore_status=True).exit_status
+            if int(exit_status) == 0:
+                self.log.debug("%s doc file exists in container" % doc)
+                test_failed = True
+        return test_failed
+
+    def test_all_nodocs(self):
+        """
+        This test checks all docs installed in container. As delivered by Platform
+        as delivered by installing packages RUN command
```
bkabrda: The part `As delivered by Platform as delivered by installing packages RUN command` doesn't make any sense to me. What were you trying to say?

phracek: Some documentation can be installed by BASE image. Like FROM ....
Nothing more.

moduleframework/tools/dockerlint.py
phracek: This is a bit tricky. We have no possibility, how to test if cache is really clean. We have to figure out Fedora and RHEL systems.

moduleframework/tools/dockerlint.py
bkabrda: Ok, I see. I don't think the quoted sentence says that, I'd recommend just removing it.

moduleframework/tools/dockerlint.py
TomasTomecek: I just made that number up in one of previous comments - when I populated cache from bigger repos, the cache had tenths of megabytes and when I did `clean all` there were still some files left. So maybe 10kB could be the threshold?

bkabrda: I haven't looked at yum, but for dnf cache, you could check that none of these files exist:

* `/var/cache/dnf/*.solv`
* `/var/cache/dnf/*.solvx`
* `/var/cache/dnf/*/metalink.xml`
* `/var/cache/dnf/*/repodata/*.xml.gz`
* `/var/cache/dnf/*/packages/*.rpm`

These are the files that seem to be deleted when you run `dnf clean all`. I'm sure you can find similar behaviour for yum. I think it'd be much better checking that none of these files exist instead of comparing directory size to a magic number that doesn't even work as I've demonstrated above.

phracek: Good point, we can separate it. If yum is used, then set threshold to 10000, in case dnf check *.solv* files. It makes more sentence then the current implementation.

bkabrda: Why would you want to use threshold for yum? Isn't it possible to use something similar to the `.solv` files for yum?

TomasTomecek: @bkabrda I tried it back then and it wasn't easy to figure out

PR #99, opened 2017-09-13
Print message for mtf-env-set

psklenar:
I partially added some user messages.
What do you more suggest to inform user about?

### DIFF
--- a/moduleframework/common.py
+++ b/moduleframework/common.py
@@ -393,14 +393,14 @@ class CommonFunctions(object):
 
         if packages:
             print_info("Installs test dependencies: ", packages)
-            # you has to be root to install some packages:
+            # you have to have root permission to install packages:
             try:
                 self.runHost(
                     "{HOSTPACKAGER} install " +
                     " ".join(packages),
                     ignore_status=False, verbose=is_debug())
             except process.CmdError as e:
-                raise CmdExc("Command failed, are you root?", e)
+                raise CmdExc("Installation failed; Do you have permission to do that?", e)
 
 
     def getPackageList(self, profile=None):
--- a/moduleframework/common.py
+++ b/moduleframework/common.py
@@ -388,15 +388,20 @@ class CommonFunctions(object):
                        will be installed.
         :return: None
         """
-
         if not packages:
             packages = self.get_test_dependencies()
+
         if packages:
             print_info("Installs test dependencies: ", packages)
-            self.runHost(
-                "{HOSTPACKAGER} install " +
-                " ".join(packages),
-                ignore_status=True, verbose=is_debug())
+            # you has to be root to install some packages:
+            try:
+                self.runHost(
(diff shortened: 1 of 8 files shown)


moduleframework/environment_prepare/docker_prepare.py
```diff
@@ -32,6 +32,7 @@ class EnvDocker(CommonFunctions):
 
     def prepare_env(self):
         self.installTestDependencies()
```
phracek: What about to document this one? I would prefer to show user, what are you installing? Does it make sense?

moduleframework/mtf_environment.py
```diff
@@ -40,8 +41,9 @@
     env = EnvNspawn()
 
 def mtfenvset():
+    print_info("Preparing environment ...")
```
phracek: Would it be possible to add also information from env.prepare_env() function? May be more details.

moduleframework/mtf_environment.py
```diff
@@ -40,8 +41,9 @@
...
     env.prepare_env()
 
 def mtfenvclean():
     env.cleanup_env()
```
phracek: The same here. More info what you are cleaning.

phracek:
Thanks for this Pull Request, though.

moduleframework/mtf_environment.py
psklenar: there is 'env.prepare_env' duplicated in backends for : EnvDocker/EnvRpm/EnvNspawn.
It throws some user info during process.

Also I changed package installation little bit, where it shows some messages what is installed and shows nice error when installation failed

moduleframework/common.py
```diff
@@ -388,15 +388,20 @@ def installTestDependencies(self, packages=None):
...
-                " ".join(packages),
-                ignore_status=True, verbose=is_debug())
+            # you has to be root to install some packages:
+            try:
+                self.runHost(
+                    "{HOSTPACKAGER} install " +
+                    " ".join(packages),
+                    ignore_status=False, verbose=is_debug())
+            except process.CmdError as e:
+                raise CmdExc("Command failed, are you root?", e)
```
jscotka: maybe better to write "do you have permissions to install packages" or something like this. because in case you are in wheel group, then you can install packages too, or solved in somehow.

moduleframework/environment_prepare/docker_prepare.py
```diff
@@ -31,6 +31,7 @@
 class EnvDocker(CommonFunctions):
 
     def prepare_env(self):
+        print_info('Loaded config for: {}'.format(self.config['module']['docker']['container']))
```
jscotka: if you have this line here, why to not have similar lines for nspawn and rpm? Could you please this to other module types too?

phracek:
LGTM.

Issue #81, opened 2017-09-06
Update documentation for MTF


phracek:
This issue should update meta-test-family.readthedocs.io with this sort of information

- [ ] How to specify various asserts in the tests
- [ ] Provide a way, every tests should be run-able standalone
- [ ]

jscotka:
Do you think that it make sense to copy UnitTest specification here.
I prefer to add link, that avocado is unittest and link to unittest docs for asserts

phracek:
Yeah, links would be enough for now.

Issue #67, opened 2017-08-25
Update Makefile examples in docs


alexxa:
[User guide page](http://10.43.2.197/~igulina/mod/build/html/user_guide/index.html) needs review and an appropriate update of points 5,6 and 8 (after PR #54 merging) according to [recent code changes](https://github.com/fedora-modularity/meta-test-family/blob/devel/examples/testing-module/Makefile).

Try to avoid explicit Makefile examples. Use links for GitHub examples files instead.

PR #100, opened 2017-09-14
[WIP] Multihost real examples

phracek:
This is one part of big #81 .
Do not merge it yet.

### DIFF
--- a/docs/user_guide/index.rst
+++ b/docs/user_guide/index.rst
@@ -24,14 +24,10 @@ User Guide
 
  .. code-block:: makefile
 
-    MODULE_LINT=/usr/share/moduleframework/tools/modulelint/*.py
-    TESTS=*.py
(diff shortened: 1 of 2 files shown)


alexxa:
@phracek Please don't copy code files in docs explicitly as for example [this one](https://github.com/fedora-modularity/meta-test-family/pull/100/files#diff-f230a59a5ae26d49f4c61c1580d5cd88R13). Since if that file will be changed for one reason or another, the docs should be updated too, and I'm afraid one will forget to do it. I believe links on files, will be more convenient. I have already mentioned it in issue #67 .

"Real Examples" sounds weird. It's not 'Real Life/World Examples' standard phrase. Just "Examples" would be ok, otherwise dictionary recommended collocations of 'example' are: typical, notable, common, illustrative, practical, concrete, particular.

phracek:
This will be solved by the big PR about documentation.

PR #113, opened 2017-09-27
remov baseruntime from Makefile

jscotka:


ignatenkobrain:
I'm wondering what's the point of sending PRs which you are immediately going to merge them... Why not to push directly?

jscotka:
Hi @ignatenkobrain, 
as we discussed with @alexxa, everything should go over PRs, but I'm working on reenable all CIs, so that I have to merge it to devel, to see results from taskotron in devel instance: 
http://taskotron-dev.fedoraproject.org/resultsdb/results?testcases=dist.modularity-testing-framework
 
So yep, I can fix it directly, but probably PRs are better trackable and maybe easier then revert them in case something wrong happen?

PR #114, opened 2017-09-27
Update dockerlint a bit according to Container:Guidelines

phracek:

Update dockerlint a bit according to Container:Guildelines
- function are named according Python coding guildelines
- new check for run or usage is added
- openshift checks are removed
- new check for summary label


jscotka:
all tests passed. Thanks a lot.

PR #125, opened 2017-10-05
New checks in dockerlinter

phracek:
Fix PR contains help.md sanity checks based on CBP

### DIFF
--- a/moduleframework/dockerlinter.py
+++ b/moduleframework/dockerlinter.py
@@ -170,7 +170,6 @@ class DockerfileLinter(object):
         :return: True if FROM is first, False if FROM is not first directive
         """
         with open(self.dockerfile) as f:
-            lines = f.readlines()
-        lines = [x for x in lines if x.isspace()]
-        lines = [x for x in lines if x.strip().startswith("#")]
+            lines = [x for x in f.readlines() if not x.isspace()]
+        lines = [x for x in lines if not x.strip().startswith("#")]
         return lines
--- a/moduleframework/dockerlinter.py
+++ b/moduleframework/dockerlinter.py
@@ -15,7 +15,8 @@ ENV = "ENV"
 PORTS = "PORTS"
 FROM = "FROM"
 RUN = "RUN"
-
+USER = "USER"
+INSTRUCT = "instruction"
 
 def get_string(value):
     return ast.literal_eval(value)
@@ -45,7 +46,7 @@ class DockerfileLinter(object):
 
     dockerfile = None
     oc_template = None
-    dfp = {}
+    dfp_structure = {}
     docker_dict = {}
 
     def __init__(self, dir_name="../"):
@@ -54,6 +55,7 @@ class DockerfileLinter(object):
             self.dockerfile = dockerfile
             with open(self.dockerfile, "r") as f:
                 self.dfp = DockerfileParser(fileobj=f)
+                self.dfp_structure = self.dfp.structure
                 self._get_structure_as_dict()
         else:
             self.dfp = None
@@ -100,13 +102,15 @@ class DockerfileLinter(object):
                      VOLUME: self._get_volume,
                      LABEL: self._get_label,
                      FROM: self._get_general,
-                     RUN: self._get_general}
+                     RUN: self._get_general,
+                     USER: self._get_general,
+                     }
 
         self.docker_dict[LABEL] = {}
         for label in self.dfp.labels:
             self.docker_dict[LABEL][label] = self.dfp.labels[label]
         for struct in self.dfp.structure:
-            key = struct["instruction"]
+            key = struct[INSTRUCT]
             val = struct["value"]
             if key != LABEL:
                 if key not in self.docker_dict:
@@ -114,8 +118,7 @@ class DockerfileLinter(object):
                 try:
                     ret_val = functions[key](val)
                     for v in ret_val:
-                        if v not in self.docker_dict[key]:
-                            self.docker_dict[key].append(v)
+                        self.docker_dict[key].append(v)
                 except KeyError:
                     print("Dockerfile tag %s is not parsed by MTF" % key)
 
@@ -164,10 +167,57 @@ class DockerfileLinter(object):
     def check_from_is_first(self):
         """
         Function checks if FROM directive is really first directive.
-        Function ignores whitespaces and comments.
         :return: True if FROM is first, False if FROM is not first directive
         """
-        with open(self.dockerfile) as f:
-            lines = [x for x in f.readlines() if not x.isspace()]
-        lines = [x for x in lines if not x.strip().startswith("#")]
-        return lines
+        if self.dfp_structure[0].get('instruction') == 'FROM':
+            return True
+        else:
+            return False
+
+    def check_from_directive_is_valid(self):
+        """
+        Function checks if FROM directive contains valid format like is specified here
+        http://docs.projectatomic.io/container-best-practices/#_line_rule_section
+        Regular expression is: ^[a-z0-9.]+(\/[a-z0-9\D.]+)+$
+        Example registry:
+            registry.fedoraproject.org/f26/etcd
+            registry.fedoraproject.org/f26/flannel
+            registry.access.redhat.com/rhscl/nginx-18-rhel7
+            registry.access.redhat.com/rhel7/rhel-tools
+            registry.access.redhat.com/rhscl/postgresql-95-rhel7
+
+        :return:
+        """
+        correct_format = False
+        struct = self.dfp_structure[0]
+        if struct.get(INSTRUCT) == 'FROM':
+            p = re.compile("^[a-z0-9.]+(\/[a-z0-9\D.]+)+$")
+            if p.search(struct.get('value')) is not None:
+                correct_format = True
+        return correct_format
+
+    def check_two_dnf_run_sections(self):
+        value = 0
+        for struct in self.dfp_structure:
+            if struct.get(INSTRUCT) == RUN:
+                value += 1
+        if int(value) > 1:
+            return False
+        return True
+
+    def check_user_number(self):
+        """
+        Function checks if user contains valid number, not string
+        :return: True if contains valid number and if it does not exit
+                 False if contains not valid number
+        """
+        try:
+            user = self.docker_dict[USER]
+        except KeyError:
+            return True
+        try:
+            for u in user:
+                int(u)
+        except ValueError:
+            return False
+        return True
--- a/moduleframework/dockerlinter.py
+++ b/moduleframework/dockerlinter.py
@@ -20,6 +20,7 @@ COPY = "COPY"
 ADD = "ADD"
 INSTRUCT = "instruction"
 
+
 def get_string(value):
     return ast.literal_eval(value)
 
@@ -200,9 +201,32 @@ class DockerfileLinter(object):
                 correct_format = True
         return correct_format
 
-    def check_chained_run_command(self):
+    def check_chained_run_dnf_commands(self):
         """
-        Function checks if Dockerfile does not contain more RUN commands in two or more rows.
+        Function checks if Dockerfile does not contain more `RUN dnf` commands
+        in more then one row.
+        BAD examples:
+             FROM fedora
+             RUN dnf install foobar1
+             RUN dnf clean all
+        GOOD example:
+             FROM fedora
+             RUN dnf install foobar1 && dnf clean all
+        :return: True if Dockerfile contains RUN dnf instructions in one row
+                False if Dockerfile contains RUN dnf instructions in more rows
+        """
+        value = 0
+        for struct in self.dfp_structure:
+            if struct.get(INSTRUCT) == RUN and "dnf" in struct.get("value"):
+                value += 1
+        if int(value) > 1:
+            return False
+        return True
+
+    def check_chained_run_rest_commands(self):
+        """
+        Function checks if Dockerfile does not contain more `RUN` commands,
+        except RUN dnf, in more then one row.
         BAD examples:
              FROM fedora
              RUN ls /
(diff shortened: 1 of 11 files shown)


moduleframework/helpfile_linter.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import print_function
+
+import os
+from moduleframework import common
+
+HELP_MD = "help.md"
+
+
+def get_help_md_file(dir_name):
+    helpmd_file = os.path.join(os.path.abspath(dir_name), "help", HELP_MD)
```
TomasTomecek: we should follow fedora guidelines here: https://fedoraproject.org/wiki/Container:Guidelines#Help_File

moduleframework/tools/dockerlint.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def setUp(self):
+        # it is not intended just for docker, but just docker packages are
+        # actually properly signed
+        self.dp = dockerlinter.DockerfileLinter()
+        if self.dp.dockerfile is None:
+            self.skip()
+
+    def test_architecture_in_env_and_label_exists(self):
+        self.assertTrue(self.dp.get_docker_specific_env("ARCH="))
```
TomasTomecek: it's not a requirement to have it as env var

moduleframework/tools/dockerlint.py
```diff
@@ -0,0 +1,112 @@
...
+class DockerLint(container_avocado_test.ContainerAvocadoTest):
+    """
+    :avocado: enable
+    """
+
+    def testBasic(self):
+        self.start()
+        self.assertTrue("bin" in self.run("ls /").stdout)
+
+    def testContainerIsRunning(self):
```
TomasTomecek: what's the purpose of this check?

moduleframework/tools/helpmd_lint.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertTrue(self.helpmd.get_tag("DESCRIPTION"))
+
+    def test_helpmd_usage(self):
+        self.assertTrue(self.helpmd.get_tag("USAGE"))
+
+    def test_helpmd_environment_variables(self):
+        self.assertTrue(self.helpmd.get_tag("ENVIRONMENT VARIABLES"))
+
+    def test_helpmd_security_implications(self):
+        self.assertTrue(self.helpmd.get_tag("SECURITY IMPLICATIONS"))
```
TomasTomecek: how is this a requirement?

moduleframework/tools/helpmd_lint.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertTrue(self.helpmd.get_tag("NAME"))
+
+    def test_helpmd_description(self):
+        self.assertTrue(self.helpmd.get_tag("DESCRIPTION"))
+
+    def test_helpmd_usage(self):
+        self.assertTrue(self.helpmd.get_tag("USAGE"))
+
+    def test_helpmd_environment_variables(self):
+        self.assertTrue(self.helpmd.get_tag("ENVIRONMENT VARIABLES"))
```
TomasTomecek: not all images operate with environment variables

requirements.txt
```diff
@@ -5,3 +5,4 @@ dockerfile-parse
 modulemd
 netifaces
 pdc-client
+mistune
```
TomasTomecek: this change seems to be unrelated to this PR

moduleframework/helpfile_linter.py
phracek: Good to know, but this should be also documented in distgen. Nowadays, we are building `help.1` file into `./root/help.1` directory and help.md itself is on help/ directory.

moduleframework/tools/dockerlint.py
phracek: It was proposed at the beginning of MTF. I guess, just for sure, container is running.
@jscotka Is it correct? Can it be removed?

moduleframework/tools/helpmd_lint.py
phracek: I will modify it, so that if container expose some ports, then the ports  should be mentioned in this label.

moduleframework/tools/dockerlint.py
jscotka: sure, just example, it just test you are in container able to do something. nothing important.

TomasTomecek: So can we remove it if it's just an example and not important?

moduleframework/dockerlinter.py
```diff
@@ -161,11 +159,15 @@ def get_specific_label(self, label_name=None):
         if label_name is None:
             return []
         label_list = self.get_docker_labels()
-        return [label_name in label_list]
+        return [label_list[key] for key in label_list.keys() if label_name == key]
 
-    def check_baseruntime(self):
+    def check_from_is_first(self):
```
TomasTomecek: Docs of this method and its implementation are two completely different things.

moduleframework/dockerlinter.py
```diff
@@ -161,11 +159,15 @@ def get_specific_label(self, label_name=None):
...
         """
-        Function returns docker labels
-        :return: label dictionary
+        Function checks if FROM directive is really first directive.
+        Function ignores whitespaces and comments.
+        :return: True if FROM is first, False if FROM is not first directive
         """
-        return [x for x in self.docker_dict.get(FROM,[]) if "baseruntime/baseruntime" in x]
+        with open(self.dockerfile) as f:
```
TomasTomecek: why parse it again when you use `DockerfileParser`?

moduleframework/helpfile_linter.py
```diff
@@ -0,0 +1,52 @@
+from __future__ import print_function
+
+import os
+from moduleframework import common
+
+HELP_MD = "help.md"
+
+
+def get_help_md_file(dockerfile):
```
TomasTomecek: still doesn't comply with Fedora Guidelines: https://fedoraproject.org/wiki/Container:Guidelines#Help_File

moduleframework/tools/dockerlint.py
```diff
@@ -0,0 +1,109 @@
...
+        if env:
+            label = self.dp.get_docker_specific_env(docker_env)
+        else:
+            label = self.dp.get_specific_label(docker_env)
+        if not label:
+            label_found = self.dp.get_specific_label(docker_label)
+        return label_found
+
+    def test_architecture_in_env_and_label_exists(self):
+        self.assertTrue(self._test_for_env_and_label("ARCH=", "architecture"))
```
TomasTomecek: Can you please point me to guidelines which require dockerfiles to have environment variable `ARCH`?

moduleframework/tools/dockerlint.py
```diff
@@ -0,0 +1,109 @@
...
+        Function tests whether labels are set in modulemd YAML file properly.
+        :return:
+        """
+        llabels = self.getConfigModule().get('labels')
+        if llabels is None or len(llabels) == 0:
+            print("No labels defined in config to check")
+            self.cancel()
+        for key in self.getConfigModule()['labels']:
+            aaa = self.checkLabel(key, self.getConfigModule()['labels'][key])
+            print(">>>>>> ", aaa, key)
```
TomasTomecek: probably a leftover from debugging

moduleframework/tools/dockerlint.py
phracek: see here: https://fedoraproject.org/wiki/Container:Guidelines#LABELS architecture is required.
But I got your point. We can only test if LABEL architecture is present.
My test was based on container-image-template: https://github.com/container-images/container-image-template/blob/master/Dockerfile.template#L5
There are mentioned as ARCH and architecture.
But I will fix it.

moduleframework/tools/dockerlint.py
```diff
@@ -0,0 +1,109 @@
...
+    #def test_from_correct_format(self):
+    #    self.assertTrue(self.dp.check_from_format())
+
+
+class DockerLint(container_avocado_test.ContainerAvocadoTest):
+    """
+    :avocado: enable
+    """
+
+    def testBasic(self):
```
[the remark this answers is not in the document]
phracek: I guess, this is some kind of sanity check. But it can be removed.

moduleframework/dockerlinter.py
phracek: Fixed.

phracek:
This Pull Request also adds checks like:
* FROM directive is valid
* USER has proper format, like only digits
* Dockerfile contains only one RUN instruction with dnf command.

moduleframework/tools/dockerlint.py
TomasTomecek: You are right, it's in there. I missed it, sorry.

Therefore we should check the presence of env vars and maybe produce a warning it they are missing.

moduleframework/dockerlinter.py
```diff
@@ -161,11 +162,62 @@ def get_specific_label(self, label_name=None):
...
+        """
+        correct_format = False
+        struct = self.dfp_structure[0]
+        if struct.get(INSTRUCT) == 'FROM':
+            p = re.compile("^[a-z0-9.]+(\/[a-z0-9\D.]+)+$")
+            if p.search(struct.get('value')) is not None:
+                correct_format = True
+        return correct_format
+
+    def check_two_dnf_run_sections(self):
```
TomasTomecek: This check does nothing with `dnf`. It should be about chained `RUN`s, so:

### bad
```
FROM fedora
RUN ls /
RUN cd /
```


### good
```
FROM fedora
RUN ls / && cd /
```

moduleframework/dockerlinter.py
```diff
@@ -161,11 +162,62 @@ def get_specific_label(self, label_name=None):
...
+        value = 0
+        for struct in self.dfp_structure:
+            if struct.get(INSTRUCT) == RUN:
+                value += 1
+        if int(value) > 1:
+            return False
+        return True
+
+    def check_user_number(self):
```
TomasTomecek: can be also string, see full spec: https://docs.docker.com/engine/reference/builder/#user

phracek: Thanks I did not find it before implementation.
That means, container maintainer should know what he is writing in Dockerfile. We have no possibility how to check this.
I will remove it.

moduleframework/dockerlinter.py
```diff
@@ -161,11 +166,72 @@ def get_specific_label(self, label_name=None):
...
+             FROM fedora
+             RUN ls / && cd /
+        :return: True if Dockerfile does not contain RUN instructions in rows
+                False if Dockerfile contains more RUN instructions in rows
+        """
+        value = 0
+        for struct in self.dfp_structure:
+            if struct.get(INSTRUCT) == RUN:
+                value += 1
+        if int(value) > 1:
```
TomasTomecek: this seems to permit only a single `RUN` instruction, that's not good

phracek: Well e.g.

~~~
FROM fedora
RUN dnf install foobar1 && dnf clean all
RUN ls /
RUN cd /
~~~
This should be just two RUN instructions.

~~~
FROM fedora
RUN dnf install foobar1
RUN dnf install clean all
RUN ls /
RUN cd /
~~~

This should be also two RUN instructions.

I would prefer again separate it into two tests. One for testing `dnf` and stuff around it and second one for testing rest instructions.

WDYT? I hate sort cuts.

TomasTomecek: Sounds like a good idea since you correctly pointed out those are 2 separate use cases.

Bear in mind that current implementation just counts `RUN`s, so e.g. [this dockerfile](https://github.com/container-images/postfix/blob/master/Dockerfile) would not pass.

moduleframework/dockerlinter.py
```diff
@@ -161,11 +167,95 @@ def get_specific_label(self, label_name=None):
...
+        """
+        correct_format = False
+        struct = self.dfp_structure[0]
+        if struct.get(INSTRUCT) == 'FROM':
+            p = re.compile("^[a-z0-9.]+(\/[a-z0-9\D.]+)+$")
+            if p.search(struct.get('value')) is not None:
+                correct_format = True
+        return correct_format
+
+    def check_chained_run_dnf_commands(self):
```
TomasTomecek: This is a good test.

moduleframework/dockerlinter.py
```diff
@@ -161,11 +167,95 @@ def get_specific_label(self, label_name=None):
...
+             RUN dnf clean all
+        GOOD example:
+             FROM fedora
+             RUN dnf install foobar1 && dnf clean all
+        :return: True if Dockerfile contains RUN dnf instructions in one row
+                False if Dockerfile contains RUN dnf instructions in more rows
+        """
+        value = 0
+        for struct in self.dfp_structure:
+            if struct.get(INSTRUCT) == RUN and "dnf" in struct.get("value"):
```
TomasTomecek: what about yum?

moduleframework/dockerlinter.py
```diff
@@ -161,11 +167,95 @@ def get_specific_label(self, label_name=None):
...
+                value += 1
+        if int(value) > 1:
+            return False
+        return True
+
+    def check_chained_run_rest_commands(self):
```
TomasTomecek: I think this test may produce some false positives: sometimes it's convenient to have multiple RUN instructions (which operate on files added to the container)

moduleframework/helpfile_linter.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import print_function
+
+import os
+from moduleframework import common
+
+HELP_MD = "help.md"
+
+
+class HelpMDLinter(object):
```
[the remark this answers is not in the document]
phracek: Good for other PR.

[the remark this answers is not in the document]
phracek: https://github.com/fedora-modularity/meta-test-family/issues/138

PR #126, opened 2017-10-05
Small fixes

jscotka:
* move default verrsion of base compose to F-27
* remove unique docker.io handling (causes tracekack in case container not defined)
* already be able to use just compose, instead of set of URLs

### DIFF
--- a/moduleframework/common.py
+++ b/moduleframework/common.py
@@ -84,12 +84,13 @@ DEFAULTRETRYTIMEOUT = 30
 DEFAULTNSPAWNTIMEOUT = 10
 
 def get_compose_url_modular_release():
-    release = os.environ.get("MTF_FEDORA_RELEASE") or "26"
+    default_release = "27"
+    release = os.environ.get("MTF_FEDORA_RELEASE") or default_release
     if release == "master":
-        release = "26"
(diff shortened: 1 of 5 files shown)


moduleframework/common.py
```diff
@@ -84,12 +84,13 @@
...
 def get_compose_url_modular_release():
-    release = os.environ.get("MTF_FEDORA_RELEASE") or "26"
+    default_release = "27"
+    release = os.environ.get("MTF_FEDORA_RELEASE") or default_release
     if release == "master":
-        release = "26"
-    compose_url = os.environ.get("MTF_BASE_COMPOSE_URL") or \
-                  "https://kojipkgs.fedoraproject.org/compose/latest-Fedora-Modular-%s/compose/Server/%s/os" \
-                  % (release, ARCH)
+        release = default_release
```
ignatenkobrain: this doesn't make any sense to me, honestly

jscotka: This is workaround around several issues. But this logic were in code for longer time. I've just improved it.
There is set of isses what I have to workaround:
* koji is slow and unreliable - we had to remove downloading all packages of dependent modules by default. 
* PDC is unrealiable, it is better to reduce PDC dependencies
* there is still missing Compose on demand service (unable to use prepared composes)

That's why some things are there now hardcoded and seems strange

PR #128, opened 2017-10-06
Small fixes2

jscotka:
fix testsuite after changes of compose url handling (already working)
revert own changes in timeout library (used cockpit original)


moduleframework/timeoutlib.py
```diff
@@ -60,9 +51,8 @@ def __enter__(self):
     def __exit__(self, *args, **kwargs):
         pass
 
-
 class Retry(object):
-    def __init__(self, attempts=1, timeout=None, exceptions=(Exception,), error=None, inverse=False, delay=None):
+    def __init__(self, attempts = 1, timeout = None, exceptions = (Exception,), error = None, inverse = False, delay = None):
```
ignatenkobrain: this doesn't seem to be according PEP8 recommendations

jscotka: I'll repair it. I've done hardcopy of https://github.com/cockpit-project/cockpit/blob/master/test/avocado/timeoutlib.py, to remove own differences

PR #133, opened 2017-10-09
systemd-run backward compotibility (without wait option)

jscotka:
experimental support for systemd-run without --wait option

### DIFF
--- a/moduleframework/helpers/nspawn_helper.py
+++ b/moduleframework/helpers/nspawn_helper.py
@@ -25,6 +25,8 @@ import re
 import glob
 import time
 import hashlib
+import string
+import random
 
 from moduleframework.common import *
 from moduleframework.exceptions import *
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
why do we need such hacks in new code?

jscotka:
@ignatenkobrain because Fedora-25 and lower versions of fedora and RedHat Enterprise Linux 7 / CentOS7
and also in taskotron production instance, default fedora version now is Fedora-25 (what caused issues for MTF automation)

But I have to do comprehensive testing of this feature. it is not intended for merging now. But comments are welcomed.

ignatenkobrain:
@jscotka f25 will become EOL in few months. talking about EL7, I would recommend you to request backporting feature on RHBZ rather than making workaround.

jscotka:
@ignatenkobrain yep, it would be better. But there is several issues:

* It will took long time to backport these options to EL7 (it is low priority, in case just me needs it)
* If nobody else  is interested in this option in EL7 it will be never fixed.

So this workaround is fast solution what will fix it  in few days instead of months or years or never. But I agree that fix the root cause sould be the best.

Otherwise I've used machinectl shell (in previous versions of MTF what was in better in some handling and working in F25 and EL-7) but it caused very ugly side-effects and I've filled more upstream bugs like (in shortcut machinectl shell is not intended to use it noniteractively :-) :
* https://github.com/systemd/systemd/issues/5879
* https://github.com/systemd/systemd/issues/5878

PR #134, opened 2017-10-10
add examples to test Red Hat and Fedora images

TomasTomecek:
Would be awesome if mtf produced good results for these images. Please run the check before merging a PR and doing a release.


TomasTomecek:
wrong branch -_-

PR #139, opened 2017-10-12
Fix some logging issues and yum checks

phracek:

This pull request fixes:
- add some log issues
- add support for yum checks

### DIFF
--- a/moduleframework/common.py
+++ b/moduleframework/common.py
@@ -53,6 +53,7 @@ PACKAGER_COMMAND = "test -e /usr/bin/dnf && echo 'dnf -y'   ||" \
 hostpackager = subprocess.check_output([PACKAGER_COMMAND], shell=True).strip()
 guestpackager = hostpackager
 ARCH = "x86_64"
+DOCKERFILE = "Dockerfile"
 
 __persistent_config = None
 
@@ -370,7 +371,6 @@ class CommonFunctions(object):
         """
         return self.info.get("url")
 
-
     def getArch(self):
         """
         Get system architecture.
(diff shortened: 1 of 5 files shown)


jscotka:
These changes seems well, it is better to move ENV handling to common module to have it on one place

moduleframework/common.py
```diff
@@ -719,3 +718,24 @@ def get_module_type_base():
...
     return parent
+
+
+def get_docker_file(dir_name="../"):
+    """
+    Function returns full path to dockerfile.
+    :param dir_name: dir_name, where should be Dockerfile located
+    :return: full_path to Dockerfile
+    """
+    fromenv = os.environ.get("DOCKERFILE")
```
TomasTomecek: this should be documented

moduleframework/common.py
```diff
@@ -719,3 +718,24 @@ def get_module_type_base():
...
+        dockerfile = fromenv
+        dir_name = os.getcwd()
+    else:
+        dir_name = os.path.abspath(dir_name)
+        dockerfile = DOCKERFILE
+    dockerfile = os.path.join(dir_name, dockerfile)
+
+    if not os.path.exists(dockerfile):
+        dockerfile = None
+        print_debug("Dockerfile should exists in the %s directory." % dir_name)
```
TomasTomecek: this should print absolute path to the directory, relative path will be confusing to users

moduleframework/common.py
TomasTomecek: http://meta-test-family.readthedocs.io/en/latest/user_guide/environment_variables.html

phracek: Documented.

TomasTomecek:
let's merge

PR #143, opened 2017-10-17
Nspanw low level library and mtf module as base module for further work

jscotka:
* nspawn moved to separated library, similar to project CONU, (but actully internally in MTF)
  * self-tests written in avocado
* add mtf module with symlinks to moduleframework


TomasTomecek:
since systemd has a python library, could be interesting to merge this into their code once proven and stable

jscotka:
Sure, that was also my idea, provide our systemd qa some tools what they can use for direct testing, and maybe add it to some systemd python api, but this is long term goal and not mine :-) Actually I wanted to have easy to debug library for nspawn operations.

jscotka:
Actually all features seems to be working well with these changes, tested locally:
with result pass.
Log attached 
[testlog.out.txt](https://github.com/fedora-modularity/meta-test-family/files/1394151/testlog.out.txt)

PR #150, opened 2017-10-19
WIP: Modulelint filtering

jscotka:
basic example how test filtering can be done with tooling (mtf tool)
see examples:
* list of tests and cout them:  `mtf -a list -l  -t tier1,fedora | wc -l`
* run tests: `mtf -l  -t fedora,tier1,optional` (it is empty, there is no testcase with tier1 and optional together)
* run just optional tests (tier1 exclueded): `mtf -l  -t optional,-tier1 `
* run all test (backward compatible): `mtf -l`

### DIFF
--- a/moduleframework/tools/__init__.py
+++ b/moduleframework/tools/__init__.py
@@ -28,6 +28,7 @@ from moduleframework import module_framework
 class Basic(module_framework.AvocadoTest):
     """
     :avocado: enable
+    :avocado: tags=docker,fedora,rhel,tier1
     """
 
     def test(self):
--- a/moduleframework/tools/helpmd_lint.py
+++ b/moduleframework/tools/helpmd_lint.py
@@ -32,6 +32,7 @@ from moduleframework.common import get_docker_file
 class HelpMDLinter(module_framework.AvocadoTest):
     """
     :avocado: enable
+    :avocado: tags=docker,fedora,rhel,optional
 
     """
 
@@ -57,11 +58,15 @@ class HelpMDLinter(module_framework.AvocadoTest):
         container_name = self.dp.get_docker_specific_env("NAME=")
         if container_name:
             self.assertTrue(self.helpmd.get_image_name(container_name[0].split('=')[1]))
+        else:
+            self.cancel()
 
     def test_helpmd_maintainer_name(self):
         maintainer_name = self.dp.get_specific_label("maintainer")
         if maintainer_name:
             self.assertTrue(self.helpmd.get_maintainer_name(maintainer_name[0]))
+        else:
+            self.cancel()
 
     def test_helpmd_name(self):
         self.assertTrue(self.helpmd.get_tag("NAME"))
@@ -82,3 +87,5 @@ class HelpMDLinter(module_framework.AvocadoTest):
     def test_helpmd_security_implications(self):
         if self.dp.get_docker_expose():
             self.assertTrue(self.helpmd.get_tag("SECURITY IMPLICATIONS"))
+        else:
+            self.cancel()
--- a/moduleframework/tools/modulelint.py
+++ b/moduleframework/tools/modulelint.py
@@ -27,8 +27,8 @@ from moduleframework import module_framework
 
 class ModuleLintSigning(module_framework.AvocadoTest):
     """
-    :avocado: disable
-    :avocado: tags=WIP
+    :avocado: enable
+    :avocado: tags=docker,fedora,rhel,tier1,WIP
     """
 
     def setUp(self):
@@ -48,12 +48,16 @@ class ModuleLintSigning(module_framework.AvocadoTest):
         for package in [x.strip() for x in allpackages.split('\n')]:
             pinfo = package.split(', ')
             if len(pinfo) == 3:
-                self.assertIn(KEY, pinfo[2])
+                if KEY in pinfo[2]:
+                    self.assertTrue(True,msg="%s in %s" % (KEY,pinfo[2]))
+                else:
+                    self.log.warn("Package sign verify failed: %s not in %s" % (KEY,pinfo[2]))
 
 
 class ModuleLintPackagesCheck(module_framework.AvocadoTest):
     """
     :avocado: enable
+    :avocado: tags=docker,fedora,rhel,tier1
     """
 
     def test(self):
--- a/tools/mtf
+++ b/tools/mtf
@@ -1,9 +1,11 @@
 #!/bin/bash
 
+AVOCADO_CMD="avocado"
 SITE_LIB=$(python -c "import moduleframework, os; print os.path.dirname(moduleframework.__file__)")
 MTF_TOOLS="tools"
-JSON_LOG=$(mktemp)
-AVOCADO_ARGS="--json $JSON_LOG"
+AVOCADO_ARGS=""
+ACTION="run"
+TAGS="--filter-by-tags-include-empty"
 
 function print_help {
   cat <<EOF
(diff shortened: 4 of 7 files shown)


moduleframework/tools/__init__.py
```diff
@@ -28,6 +28,7 @@
 class Basic(module_framework.AvocadoTest):
     """
     :avocado: enable
+    :avocado: tags=docker,fedora,rhel,tier1
     """
 
     def test(self):
```
phracek: This could be a bit complicated. We are calling avocado run *.py in tools directory and __init__.py is executed as well.
What is this test used for? I don't see any reason

moduleframework/tools/helpmd_lint.py
```diff
@@ -57,11 +58,15 @@ def test_helpmd_image_name(self):
         container_name = self.dp.get_docker_specific_env("NAME=")
         if container_name:
             self.assertTrue(self.helpmd.get_image_name(container_name[0].split('=')[1]))
+        else:
+            self.cancel()
```
phracek: Why cancel? If container_name does not exist, then it should fail.

moduleframework/tools/modulelint.py
```diff
@@ -27,8 +27,8 @@
 
 class ModuleLintSigning(module_framework.AvocadoTest):
     """
-    :avocado: disable
-    :avocado: tags=WIP
+    :avocado: enable
+    :avocado: tags=docker,fedora,rhel,tier1,WIP
```
phracek: We should think how to structure tests. Some tests are mandatory and some tests are optional.
This should be discussed with other folks like Alex Corwin.

tools/mtf
```diff
@@ -13,8 +15,39 @@ positional arguments:
   AVOCADO_ARGS        test files and any arguments that "avocado run" accepts
 
 optional arguments:
+  -a, --action        run selected avocado command (DEFAULT: run)
   -h, --help          show this help message and exit
   -l, --linters       also run default linter tests included in MTF
+  -t, --tags T1,-T2   set test filtering based on tags, use --taglist for help ()
```
phracek: remote (-) before T2

tools/mtf
```diff
@@ -13,8 +15,39 @@ positional arguments:
...
+function print_tags {
+  cat <<EOF
+list of well known tags:
+  test environments:
+    rhel     - working on RedHat Enterprise Linux and CentOs
+    fedora   - working on Fedora
+    docker   - working on Docker
+
+  test categorization:
+    tier1    - basic qualification tests
```
phracek: I would name it differently. tier1 does not sound well. My opinion.

tools/mtf
```diff
@@ -27,16 +60,36 @@ while [[ $# -gt 0 ]]; do
...
+
+# append filtering
+test -n "$TAGS"        && AVOCADO_ARGS="$AVOCADO_ARGS $TAGS"
+
 # preserve exist status, mtf command should finish same as testing.
-avocado run $AVOCADO_ARGS
+echo "RUNNING: $AVOCADO_CMD $ACTION $AVOCADO_ARGS"
+$AVOCADO_CMD "$ACTION" $AVOCADO_ARGS
 EXIT_STATUS=$?
-mtf-log-parser $JSON_LOG
```
phracek: Why you removed mtf-log-parser? Any reason? Can you please paste how does it look like output now?

tools/mtf
```diff
@@ -13,8 +15,39 @@ positional arguments:
...
+                      http://avocado-framework.readthedocs.io/en/55.0/Loaders.html#filtering-tests-by-tags
+
+EOF
+}
+
+function print_tags {
+  cat <<EOF
+list of well known tags:
+  test environments:
+    rhel     - working on RedHat Enterprise Linux and CentOs
```
TomasTomecek: it's 'Red Hat E...' and 'CentOS'

tools/mtf
accorvin: Perhaps "sanity" would be a good name?

tools/mtf
```diff
@@ -13,8 +15,39 @@ positional arguments:
...
+  test categorization:
+    tier1    - basic qualification tests
+    tier2    - extended qualification tests
+    optional - could fail on some variants (not everything is supported)
+    slow     - slow testcases, longer than expected (hrs, or days)
+    WIP      - work in progress tests, could be bad, failing or under developlment
+
+In tests use docs string like (to allow run test on fedora, rhel and docker env and it is tier1 test).
+It is possible to use tags in doc string in class or in class method.
+Methods tags are ingerited from class tags
```
accorvin: typo in "inherited"

moduleframework/tools/modulelint.py
accorvin: I agree that this is the most important feature for me. Some container owners are going to have valid reasons to want to disable a specific test. Ideally, that container owner could keep some sort of metadata file alongside their container/dockerfile that would specify which tests to enable/disable. I'm thinking something similar to the pylintrc for the pylint python linter. When you run the pylint tool, it knows to look for a pylintrc file that controls what tests to skip.

moduleframework/tools/__init__.py
jscotka: I don't need it, this is caused by moving module linters from /usr/share/module../modulelit -> sitepackages. This is important for installation via pip an, in case, in case there is no `init` file. pip install creates empty one and this causes that run linters via *.py causes traceback, because python file is empty. 
Thats why at the begining I were against moving modulelinters to sitepackages from some well known location.

moduleframework/tools/helpmd_lint.py
jscotka: But it does not fails anyhow and mark test as PASS, although this testcase does not test anything and this is very bad.
This improves behavior, at least that it is now canceled, in case "if container_name" is not true instead of PASS.

moduleframework/tools/modulelint.py
jscotka: Yep, we have to think about that more. what it the best way how to do that. This PR uses avocado tags, as main filter possibilities, so everything is possible to filter from `cmdline`, but maybe we can define it also in some file what `mtf` tool will use, instead of `-t` param from cmdline. I can imagine something like file `.mtfrc.yaml` with content like:
```
module_lint_tags:
     - sanity,-optional
     - composecheck
```
what will be transformed to `mtf` args `-t sanity,-optional -t composecheck`. Does it make sense?

tools/mtf
jscotka: It is intended there. It as example that you can use __negation__ of filter  meaning is T1 AND (NOT T2) it is semantic provided by avocado.

tools/mtf
jscotka: I'm not against, but there is more troubles with __sanity__
there is more levels of testing: like unittest, component tests (smoke, sanity, regressions), integrations tests, compose testing and much more. 
these levels and sometimes this can cause misunderstanding on both sides :-)
So ther is possible have tags like: `tier0, tier1, tier2, tier3` ... or have tags like `unittest, smoke, sanity, regression, integration, ...` 
* __tierX__ - not descriptive, but easy to understand that there is more levels, and that levels has some structure
* __sanity,... - descriptive, but not easy to see structure of levels

I'm not against any solution. or maybe we can tag these tests by both tags, `tier1` and `sanity`, it allows to use both names

tools/mtf
jscotka: it is not removed, just moved to another line, see code, because this log parser is intended just when you use default action `run` in case for example you will call `-a list` it leads to bad code

moduleframework/tools/modulelint.py
accorvin: Yes, I think that sounds perfect.

phracek: .pylintrc has a INI file like format. Do you need really YAML structure?
Based on the output from `pylint --generate-rcfile`

accorvin: Nope, I don't think the exact format matters

jscotka:
it is replaced by
https://github.com/fedora-modularity/meta-test-family/pull/165
there is full metadata parser

PR #154, opened 2017-10-24
Fix help issue

phracek:
This issue fixes two problems.

* If help.md is not present in Dockerfile, then exit as WARN
* if help.md is not present in the same directory as Dockerfile, then skip al helpmd tests


moduleframework/tools/dockerlint.py
```diff
@@ -94,7 +94,10 @@ def test_chained_run_rest_commands(self):
         self.assertTrue(self.dp.check_chained_run_rest_commands())
 
     def test_helpmd_is_present(self):
-        self.assertTrue(self.dp.check_helpmd_is_present())
+        helpmd_present = self.dp.check_helpmd_is_present()
+        if not helpmd_present:
+            self.log.warn("help.md is not present in Dockerfile")
+        self.assertTrue(True)
```
TomasTomecek: I don't know avocado, but is this really needed? Does every test need to be asserted?

phracek: I have switched the test from FAIL and PASS to PASS and WARN, because of help.md is not important.
Otherwise I will revert it.

jscotka: This `self.assertTrue(True)` is very bad from my PoV,  maybe easy to understand could be with comment:
```
if not helpmd_present:
    # you have to have at lease one assert to report state as __WARN__
    self.assertTrue(True) 
    self.log.warn("help.md is not present in Dockerfile")
else:
    self.assertTrue(helpmd_present)
```
or maybe even better create and add some decorator around self.assert*

like
```
class X...
    def mark_as_warn(self,function,*args, **kwargs):
       try:
          function(*args, **kwargs)
       except Exception as e:
          self.assertTrue(True)
          self.log.warn("Warning raised: %s" % e)

and then use inside test like: ` self.mark_as_warn(self.assertTrue, False)`

what do you think? We can add this to AvocadoTest class https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/avocado_testers/avocado_test.py#L40

and then you can use this function everywhere you will use assert but expect to return just warn

Comment (TomasTomecek) [reply moduleframework/tools/dockerlint.py]: seems like a bug in avocado (that you need to assert in order to make a warn)

Comment (jscotka) [reply moduleframework/tools/dockerlint.py]: I've tried that and it is not important to have there this assertTrue(True) anyhow.
@phracek you can have test without any assert and it is possible to print warn well, it also means that fixes like https://github.com/fedora-modularity/meta-test-family/pull/150/files#diff-89b2554ab29801f3d6668f1e0fa567f9R62 are important, because now, these tests pass, although they do nothing, no asserts means pass by default, what is expected behavior.

```
from avocado import Test

class X(Test):

    def testNo(self):
        pass

    def testTrue(self):
        self.assertTrue(True)

    def testFalse(self):
        self.assertTrue(False)

    def testWarn(self):
        self.log.warn("pass but bad")
```
returns:
```
 (1/4) test.py:X.testNo: PASS (0.01 s)
 (2/4) test.py:X.testTrue: PASS (0.02 s)
 (3/4) test.py:X.testFalse: FAIL (0.04 s)
 (4/4) test.py:X.testWarn: WARN (0.00 s)
```
so your code could be writen like: 
```
if not helpmd_present:
    self.log.warn("help.md is not present in Dockerfile")
else:
    self.assertTrue(helpmd_present)
```

or write some generic method and use it to transform assert to warn and use it then as described:
```
def mark_as_warn(self,function,*args, **kwargs):
       try:
          function(*args, **kwargs)
       except AssertionError as e:
          self.log.warn("Warning raised: %s" % e)
```

**Comment by phracek** [comment]
All issues were addressed.

File: moduleframework/avocado_testers/avocado_test.py [hunk 2]
```diff
@@ -263,6 +263,19 @@ def run_script(self, *args, **kwargs):
         """
         return self.backend.run_script(*args, **kwargs)
 
+    def mark_as_warn(self, func, *args, **kwargs):
+        """
+        run function which you would like to mark as WARN
+        :param func: function for run
+        :param args: pass this args to run function
+        :param kwargs: pass thru to avocado process.run
```
jscotka: This is not true, this is passed same as args as key-value to `func`, not to avocado process run.

moduleframework/avocado_testers/avocado_test.py
```diff
@@ -263,6 +263,19 @@ def run_script(self, *args, **kwargs):
         """
         return self.backend.run_script(*args, **kwargs)
 
+    def mark_as_warn(self, func, *args, **kwargs):
```
jscotka: maybe would be better call this functin, `assert_to_warn` to explicitly say, that you are transforming assert, not any exception. I know that I've also choosed this name, I were lazy to invent some better one :-)

moduleframework/helpfile_linter.py
```diff
@@ -49,7 +49,10 @@ def get_maintainer_name(self, name):
 
     def get_tag(self, name):
         name = '# %s' % name
+        tag_found = True
         if not self.help_md:
+            common.print_info("help md does not exist.")
             return False
-        tag_exists = [x for x in self.help_md if name.upper() in x]
-        return tag_exists
+        if not [x for x in self.help_md if name.upper() in x]:
```
jscotka: I'm not sure with this statement, what you have there means?
*  `if name.upper() not in self.help_md`
* or it means that you are searching for substring, in that case you can use something like: `if name.upper() not in " ".join(self.help_md)`, in case you are not searching for substring this and original code will lead to mistake, when for example name.upper() == "SOME" and x will contain for example "SOMEONE" 
* or it means that you are searching by `if name.upper() in x` in more complex object? like dict, in this case this usage is okay.

phracek: It means, go through help_md file line by line. Search for given string in line somewhere and before searching it convert to upper.

jscotka: thanks for explanation. So it make sense to have it in this form and equivalent is something like: `if name.upper() not in "\n".join(self.help_md)`

Issue #147, opened 2017-10-18
[RFE] General test suite for dockerfiles and images needs to be configurable


phracek:
We need to figure out a way, how to make the test run configurable.
AC:
- [ ] users are able to configure a specific test case whether its bad result can be ignored
- [ ] users are able to configure test cases (e.g. whitelists for `rpm -`qd` checks) - the UI should be discussed with Alex
- [ ] document the new functionality

jscotka:
# Solved by
https://github.com/fedora-modularity/meta-test-family/pull/150

Internal discussion, then we update state here

TomasTomecek:
@jscotka can you please elaborate how is this issue solved by the linked PR? I feel like the linked PR is just about filtering and which tests should be run; the point 1 in this issue is about running all tests and just "waiving" their results

phracek:
This is already solved by #150.

PR #158, opened 2017-10-30
Split linters into more classes

phracek:

This Pull request splits DockerFileLinter into more classes so we are able to specify which tests can be run.
It covers a bit issues #147

### DIFF
--- a/moduleframework/tools/dockerlint.py
+++ b/moduleframework/tools/dockerlint.py
@@ -28,9 +28,44 @@ from moduleframework import dockerlinter
 from moduleframework.avocado_testers import container_avocado_test
 
 
-class DockerFileLinter(module_framework.AvocadoTest):
+class DockerInstructions(module_framework.AvocadoTest):
     """
     :avocado: enable
+    :avocado: tags=sanity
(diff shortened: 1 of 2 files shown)


moduleframework/tools/dockerlint.py
```diff
@@ -28,9 +28,44 @@
 from moduleframework.avocado_testers import container_avocado_test
 
 
-class DockerFileLinter(module_framework.AvocadoTest):
+class DockerInstructions(module_framework.AvocadoTest):
```
TomasTomecek: `DockerfileInstructionsTests` maybe?

moduleframework/tools/helpmd_lint.py
```diff
@@ -29,9 +29,10 @@
 from moduleframework.common import get_docker_file
 
 
-class HelpMDLinter(module_framework.AvocadoTest):
+class DockerHelpSanity(module_framework.AvocadoTest):
```
TomasTomecek: I would put `HelpFile` in it, the present name is a bit confusing.

PR #161, opened 2017-10-31
Make function simple.

phracek:

This fixes function assert_to_warn which causes troubles in travis and container-images-template


phracek:
Does not make sense

PR #162, opened 2017-11-02
rebase of mtf cmd line

psklenar:


moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+import tempfile
+import shutil
+
+from sys import argv
+
+from avocado.utils import process
+from avocado.core.app import AvocadoApp
+from avocado.core.parser import Parser
+
+logger = logging.getLogger("mtf")
```
phracek: I would prefer to have it in common.py so we can use it in whole MTF https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/common.py

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def set_logging(level=logging.INFO):
```
phracek: The same this.

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+
+    set_logging(level=logging.DEBUG if args.verbose else logging.INFO)
+
+    logger.debug("Options: verbose={0}, name={1}, linter={2}, setup={3}, action={4}, module={5}".format(
+        args.verbose, args.name, args.linter, args.setup, args.action, args.module))
+    logger.debug("remaining options for avocado: {0}".format(unknown))
+
+    # environment
+    # hint for read: os.environ.get('MODULE') ; write os.environ['MODULE']
+    if os.environ.get('MODULE') is None:
```
phracek: I guess, if MODULE is not specified, than it should failed. It is important, isn't it?

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+            else:
+                # this is additional avocado param
+                self.additionalAvocadoArg += " {0} ".format(param)
+
+        logger.debug("tests = {0}".format(self.tests))
+        logger.debug("additionalAvocadoArg = {0}".format(self.additionalAvocadoArg))
+
+
+    def bash_command(self, cmd):
+         return process.run(cmd, shell=True, ignore_status=True)
```
phracek: We have self.runHost in MTF. I would prefer to remove it and makes it general. https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/common.py#L400

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,163 @@
...
+        if os.path.exists(self.tmppath):
+            cmd = self.bash_command("mtf-log-parser {JSON_LOG}".format(JSON_LOG=self.tmppath))
+            logger.info(cmd.stdout)
+            logger.debug("cmd was {0}".format(cmd.command))
+            os.remove(self.tmppath)
+
+def main():
+    args, unknown = cli()
+
+    if args.setup:
```
phracek: Is it plan to remove mtf-env-setup? I don't understant, why it is part of `mtf`

moduleframework/mtf_scheduler.py
psklenar: I use common.print_debug + common.print_info,
(probably) I will create new issue request to modify common.py with better logging ability for these methods

moduleframework/mtf_scheduler.py
psklenar: I use common.print_debug + common.print_info,
(probably) I will create new issue request to modify common.py with better logging ability for these methods

moduleframework/mtf_scheduler.py
psklenar: NOW  MODULE could be from:
    #   1. common.get_module_type() ... it reads config.yaml and treceback if it doesn't exist
    #   2. environment ... MODULE=docker etc
    #   3. argument ... --module=docker

moduleframework/mtf_scheduler.py
psklenar: self.runHost is not usable as it load config (which cannot exists) and after discussion with  @jscotka I should use own stuff in this case

moduleframework/mtf_scheduler.py
psklenar: I add -s / --setup  to mtf, default is 'false' and yes it calls 'mtf-env-setup'
I feel it could do this setup, or not?

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+    # environment usage:
+    #   read: os.environ.get('MODULE')
+    #   write: os.environ['MODULE'] 
+
+    # MODULE could be from:
+    #   1. common.get_module_type() ... it reads config.yaml and treceback if it doesn't exist
+    #   2. environment ... MODULE=docker etc
+    #   3. argument ... --module=docker
+    try:
+        args.module=common.get_module_type()
```
jscotka: please check your code and repair all troubles like this
there has to be `A = B`

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+        # TODO it wrongly writes: 'Using default minimal config ...', change in common.py
+    except moduleframework.mtfexceptions.ModuleFrameworkException:
+        pass
+
+    if os.environ.get('MODULE') is not None:
+        # environment is the highest priority because mtf uses environment (too much)
+        args.module=os.environ['MODULE']
```
jscotka: like this

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+    )
+    parser.add_argument("--linter", "-l", action="store_true", default=False, help='adds additional compose checks')
+    parser.add_argument("--setup", action="store_true", default=False, help='Setup by mtfenvset')
+    parser.add_argument("--action", action="store", default='run', help='Action for avocado, see avocado --help for subcommands')
+
+    # some useful bash variables
+    # there are more possible variables up to the doc, not sure what could be like options too
+    group = parser.add_argument_group('additional arguments are like environment variables up to the http://meta-test-family.readthedocs.io/en/latest/user_guide/environment_variables.html ')
+    group.add_argument("--module", action="store", help='Module type, like: docker, nspawn or rpm')
+    group.add_argument("--debug", action="store_true", help='passes AVOCADO_LOG_DEBUG=yes DEBUG=yes')
```
jscotka: Please improve help messages, not just test that it set these envvars, some high level function of this option

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+    return args, unknown
+
+class AvocadoStart(object):
+    def __init__(self, args, unknown):
+
+        self.module = os.environ['MODULE']
+        self.action = args.action
+
+        # its used for filepath in loadCli:
+        self.tmppath = tempfile.mktemp()
```
jscotka: this creates file everytime and in case you'll use another action like `list` it causes that it create file in constructor and does not remove this file anywhere, it is connected trouble with `def avocado_run` ,
name  and login in main invocate that it will be called just in case `avocado run` command, but it takes  `self.action` param, so that it is possible to invoke it with another action -> it causes troubles with this file, because `list` action does not allow to pass --json..* params or similar

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+        # run avocado with right cmd arguments
+        # bash = self.bash_command(("{AVOCADO} {a} {b}".format(AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=self.tests)))
+        bash = self.bash_command(("{AVOCADO} {a} {b}".format(AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=self.tests)))
+        self.returncode += bash.exit_status
+        common.print_info(bash.stdout, bash.stderr)
+
+    def show_error(self):
+        # show errors details and delete log file
+        if os.path.exists(self.tmppath):
+            bash = self.bash_command("mtf-log-parser {JSON_LOG}".format(JSON_LOG=self.tmppath))
```
jscotka: please rename `self.tmppath` to for example `self.json_tmppath` to be clear what is this temp file, because it is instance varibale
and second important thing is, why you call it via __bash command__ it is python inside MTF, so this does not make sense to call it via bash UI for this python

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+            common.print_info(bash.stdout, bash.stderr)
```
jscotka: I'm not sure if this is good, it causes something like bash "echo x >&2 " in final, so everything is redirected to stderr. 
maybe would be good to let stderr on stderr and stdout on stdout -> do not use `common.print_info` function here, and use here own print here to be able to split outputs

Otherwise this is exactly reason why you had add 2>&1 to Makefile, because previous `mtf` commands output everything to stdout

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+    common.print_debug('verbose/debug mode')
+    args, unknown = cli()
+
+    if args.setup:
+        # mtfenvset need bash environment!
+        from moduleframework.mtf_environment import mtfenvset
+        mtfenvset()
+
+    a = AvocadoStart(args, unknown)
+    if args.action == 'run':
```
jscotka: so any other avocado action is not calle, although in avocado_run it use param `self.action`

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def main():
+    common.print_debug('verbose/debug mode')
+    args, unknown = cli()
+
+    if args.setup:
+        # mtfenvset need bash environment!
+        from moduleframework.mtf_environment import mtfenvset
+        mtfenvset()
```
jscotka: this leads to potential issues, `mtfenvset()` needs to have also set up __env variables__ and this part is done later in `a = AvocadoStart()` it causes troubles probably. for example if you will use `MODULE=docker mtf --module=nspawn ...` it leads that `mtfenvset()` will be invoked with `docker` and `avocado run` will be invoked with `nspawn`

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def avocado_run(self):
+        avocado_args="--json {JSON_LOG}".format(JSON_LOG=self.tmppath)
+        avocadoAction = "MODULE={MODULE} avocado {ACTION} {AVOCADO_ARGS}".format(MODULE=self.module, ACTION=self.action, AVOCADO_ARGS=avocado_args)
+
+        # run avocado with right cmd arguments
+        # bash = self.bash_command(("{AVOCADO} {a} {b}".format(AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=self.tests)))
+        bash = self.bash_command(("{AVOCADO} {a} {b}".format(AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=self.tests)))
+        self.returncode += bash.exit_status
+        common.print_info(bash.stdout, bash.stderr)
```
jscotka: you have this twice time, herre and in `show_error`

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+                # this is additional avocado param
+                self.additionalAvocadoArg += " {0} ".format(param)
+
+        common.print_debug("tests = {0}".format(self.tests))
+        common.print_debug("additionalAvocadoArg = {0}".format(self.additionalAvocadoArg))
+
+    def bash_command(self, cmd):
+        # only to run bash command
+        common.print_debug("cmd: {0}".format(cmd))
+        return process.run(cmd, shell=True, ignore_status=True)
```
jscotka: Why there is `ignore_status=True` as default. in case for example some command isn't found, can cause that it leads to __false negative__ , better is to ensure that all commands pass, and in case you expect that it fails because of some reason you should check return code. so this make sense to ignore this probably just with "avocado run" command what then report exit status to last command like `exit(RC)`

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def avocado_run(self):
+        avocado_args="--json {JSON_LOG}".format(JSON_LOG=self.tmppath)
+        avocadoAction = "MODULE={MODULE} avocado {ACTION} {AVOCADO_ARGS}".format(MODULE=self.module, ACTION=self.action, AVOCADO_ARGS=avocado_args)
+
+        # run avocado with right cmd arguments
+        # bash = self.bash_command(("{AVOCADO} {a} {b}".format(AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=self.tests)))
+        bash = self.bash_command(("{AVOCADO} {a} {b}".format(AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=self.tests)))
+        self.returncode += bash.exit_status
```
jscotka: why this `+=` ? I don't undestart that logic, it meand that there will be more `avocado_run` commands and sum of output code means something? I thing that just `return bash.exit_status` could be enough and you don't need this instance variable `self.returncode` anyhow

moduleframework/mtf_scheduler.py
psklenar: done

moduleframework/mtf_scheduler.py
psklenar: done

moduleframework/mtf_scheduler.py
psklenar: the whole tool is designed mainly for avocado run, which can return non zero return code. Such a results is parsed by next tool for logging errors, thus ignore_status=True is needed.

moduleframework/mtf_scheduler.py
psklenar: yes, its intentional, together with running command with ignore_status=True you know what all fails.

I can return  just one stable return code for different method (avocado run ~ return code X, parsing error return code Y etc.) - but I felt that more higher return code indicates bigger issue, that's why that += ; should I really change it?

moduleframework/mtf_scheduler.py
psklenar: done

moduleframework/mtf_scheduler.py
psklenar: it was changed

moduleframework/mtf_scheduler.py
psklenar: in this case it will be for MODULE=docker, as all the cli params are forwarded as bash variables by design. Bash variables control this whole project so I decided to make it higher priority.

user could see which module is used in '--debug' mode

moduleframework/mtf_scheduler.py
psklenar: done

moduleframework/mtf_scheduler.py
psklenar: done, autopep8 -i

moduleframework/mtf_scheduler.py
psklenar: done, autopep8 -i

moduleframework/mtf_scheduler.py
jscotka: yep, it make sense, so it case you can remove this method `bash_command` at all, and use `process.run("avocado what you want", ignore_status=True)` directly in your code, this method is just simple wrapper around `process.run` (btw, you can remove `shell=True` from that) if you don't want to call some bash specific things, (eg. `shell=True` makes sense in case you would like to use command with pipes or use some expansions like `avocado ... *.py ` or `avocado run ... | grep output` what is not your case)

moduleframework/mtf_scheduler.py
jscotka: Yep, that make sense and it is proper way, I tried to check the logic inside code, now I found, that you have this logic in `cli()` code what does parsing of cmd line parameters.

phracek:
LGTM

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,214 @@
...
+            AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=self.tests), shell=True, ignore_status=True)
+        common.print_info(bash.stdout, bash.stderr)
+        return bash.exit_status
+
+    def avocado_general(self):
+        # additional parameters
+        # self.additionalAvocadoArg: its from cmd line, whats unknown to this tool
+        avocado_args = ""  # when needed => specify HERE your additional stuff
+        avocadoAction = "avocado {ACTION} {AVOCADO_ARGS}".format(
+            MODULE=self.module, ACTION=self.action, AVOCADO_ARGS=avocado_args)
```
jscotka: remove please `MODULE=` not used here

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,214 @@
...
+        bash = process.run("{AVOCADO} {a} {b}".format(
+            AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=self.tests), shell=True, ignore_status=True)
```
jscotka: why you have here `shell=True` ? does that any sense?

moduleframework/mtf_scheduler.py
```diff
@@ -0,0 +1,214 @@
...
+        if os.path.exists(self.json_tmppath):
+            try:
+                # file has to by json, otherwise it fails
+                json_data = open(self.json_tmppath).read()
+                data = json.loads(json_data)
+            except (IOError, ValueError) as e:
+                # file is not readable as json: No JSON object could be decoded
+                print(e)
+                # fatal error when this command fails, its unexpected
+                exit(127)
```
jscotka: please remove this file `self.json_tmppath` if error appear causes that it let there this file, because of exit.

moduleframework/mtf_scheduler.py
psklenar: done

moduleframework/mtf_scheduler.py
psklenar: removed, done

moduleframework/mtf_scheduler.py
psklenar: ok, done

jscotka:
We can merge it. I'm happy with these changes/].

Issue #89, opened 2017-09-07
[RFE] Support testing containers in OpenShift


phracek:
MTF should be able to support testing container in OpenShift environment.
We'd have list of templates and checks, framework runs openshift somehow (or connects to existing one), runs sanity check for the templates, deploy the templates, and run the probe that is specified for every template.

- [x] MTF will call it like MODULE=openshift
- [x] MTF is able to connect to an OpenShift environment, specified by OPENSHIFT_IP variable otherwise use `localhost`
- [x] MTF calls `new-app` for creating an application in the OpenShift environment, based on the docker image
- [x] MTF deletes a new-app from OpenShift environment
- [x] MTF checks whether PORT is really opened. Testing example could be `memcached`
- [x] Provide a clear documentation

phracek:
This is already covered by #164.

PR #164, opened 2017-11-08
testing container in OpenShift

phracek:
This Pull Request brings possibility to test containers in OpenShift environment.

MTF provides two scripts for preparing environment like `MODULE=openshift mtf-env-set` and `MODULE=openshift mtf-env-clean`

Usage is like:
`MODULE=openshift OPENSHIFT_IP="Openshift_IP_Address" OPENSHIFT_USERNAME=developer OPENSHIFT_PASSWD=developer mtf -l *.py`

The Pull Request fixes issue #89

### DIFF
--- a/docs/user_guide/environment_setup.rst
+++ b/docs/user_guide/environment_setup.rst
@@ -24,7 +24,7 @@ Manual Setup
 **OpenShift**
 
  - Install OpenShift if not installed and if environment variable ``OPENSHIFT_LOCAL`` is specified.
- - if ``OPENSHIFT_LOCAL`` variable is specified, the it starts an OpenShift by command ``oc cluster up`` or stops it by command ``oc cluster down``.
+ - if ``OPENSHIFT_LOCAL`` variable is specified, then it starts an OpenShift by command ``oc cluster up`` or stops it by command ``oc cluster down``.
 
 Automated Setup
 ~~~~~~~~~~~~~~~
--- a/docs/user_guide/environment_setup.rst
+++ b/docs/user_guide/environment_setup.rst
@@ -21,13 +21,18 @@ Manual Setup
 
  - No any configuration needed
 
+**OpenShift**
+
+ - Install OpenShift if not installed and if environment variable ``OPENSHIFT_LOCAL`` is specified.
+ - if ``OPENSHIFT_LOCAL`` variable is specified, the it starts an OpenShift by command ``oc cluster up`` or stops it by command ``oc cluster down``.
+
 Automated Setup
 ~~~~~~~~~~~~~~~
 
 The environment configuration scripts should be executed in the same directory where the tests are, otherwise the environment variable **CONFIG** should be set.
 
   - to setup environment run ``MODULE=docker mtf-env-set``
-  - to execute tests run ``MODULE=docker avocado run your.test.py``
+  - to execute tests run ``MODULE=docker mtf your.test.py``
   - to cleanup environment ``MODULE=docker mtf-env-clean``
 
 Test Creation
--- a/docs/user_guide/environment_variables.rst
+++ b/docs/user_guide/environment_variables.rst
@@ -25,10 +25,10 @@ Environment variables allow to overwrite some values of a module configuration f
 - **MTF_REMOTE_REPOS=yes** disables downloading of Koji packages and creating a local repo, and speeds up test execution.
 - **MTF_DISABLE_MODULE=yes** disables module handling to use nonmodular test mode (see `multihost tests`_ as an example).
 - **DOCKERFILE="<path_to_dockerfile"** overwrites the location of a Dockerfile.
-- **OPENSHIFT_LOCAL** enables installing ``origin`` and ``origin-clients`` on local machine
-- **OPENSHIFT_IP** uses this IP address for connecting to an OpenShift environment.
-- **OPENSHIFT_USER** uses this ``USER`` name for login to an OpenShift environment.
-- **OPENSHIFT_PWD** uses this ``PWD`` name for login to an OpenShift environment.
+- **OPENSHIFT_LOCAL=yes** enables installing ``origin`` and ``origin-clients`` on local machine
+- **OPENSHIFT_IP=openshift_ip_address** uses this IP address for connecting to an OpenShift environment.
+- **OPENSHIFT_USER=developer** uses this ``USER`` name for login to an OpenShift environment.
+- **OPENSHIFT_PASSWORD=developer** uses this ``PASSWORD`` name for login to an OpenShift environment.
 
 .. _multihost tests: https://github.com/fedora-modularity/meta-test-family/tree/devel/examples/multios_testing
 
--- a/docs/user_guide/environment_variables.rst
+++ b/docs/user_guide/environment_variables.rst
@@ -25,6 +25,10 @@ Environment variables allow to overwrite some values of a module configuration f
 - **MTF_REMOTE_REPOS=yes** disables downloading of Koji packages and creating a local repo, and speeds up test execution.
 - **MTF_DISABLE_MODULE=yes** disables module handling to use nonmodular test mode (see `multihost tests`_ as an example).
 - **DOCKERFILE="<path_to_dockerfile"** overwrites the location of a Dockerfile.
+- **OPENSHIFT_LOCAL** enables installing ``origin`` and ``origin-clients`` on local machine
+- **OPENSHIFT_IP** uses this IP address for connecting to an OpenShift environment.
+- **OPENSHIFT_USER** uses this ``USER`` name for login to an OpenShift environment.
+- **OPENSHIFT_PWD** uses this ``PWD`` name for login to an OpenShift environment.
 
 .. _multihost tests: https://github.com/fedora-modularity/meta-test-family/tree/devel/examples/multios_testing
 
--- a/moduleframework/avocado_testers/openshift_avocado_test.py
+++ b/moduleframework/avocado_testers/openshift_avocado_test.py
@@ -27,7 +27,7 @@ from moduleframework.common import get_module_type_base
 # INTERFACE CLASSES FOR SPECIFIC MODULE TESTS
 class OpenShiftAvocadoTest(AvocadoTest):
     """
-    Class for writing tests specific just for DOCKER
+    Class for writing tests specific just for OpenShift
     derived from AvocadoTest class.
 
     :avocado: disable
--- a/moduleframework/avocado_testers/openshift_avocado_test.py
+++ b/moduleframework/avocado_testers/openshift_avocado_test.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+#
+# This Modularity Testing Framework helps you to write tests for modules
+# Copyright (C) 2017 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# he Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Petr Hracek <phracek@redhat.com>
+#
+
+from moduleframework.module_framework import AvocadoTest
+from moduleframework.common import get_module_type_base
+
+
+# INTERFACE CLASSES FOR SPECIFIC MODULE TESTS
+class OpenShiftAvocadoTest(AvocadoTest):
+    """
+    Class for writing tests specific just for DOCKER
+    derived from AvocadoTest class.
+
+    :avocado: disable
+    """
+
+    def setUp(self):
+        if get_module_type_base() != "openshift":
+            self.cancel("OpenShift specific test")
+        super(OpenShiftAvocadoTest, self).setUp()
+
+    def checkLabel(self, key, value):
+        """
+        check label of docker image, expect key value (could be read from config file)
+
+        :param key: str
+        :param value: str
+        :return: bool
+        """
+        if key in self.backend.containerInfo['Labels'] and (
+                    value in self.backend.containerInfo['Labels'][key]):
+            return True
+        return False
+
+
--- a/moduleframework/avocado_testers/openshift_avocado_test.py
+++ b/moduleframework/avocado_testers/openshift_avocado_test.py
@@ -37,18 +37,3 @@ class OpenShiftAvocadoTest(AvocadoTest):
         if get_module_type_base() != "openshift":
             self.cancel("OpenShift specific test")
         super(OpenShiftAvocadoTest, self).setUp()
-
-    def checkLabel(self, key, value):
-        """
-        check label of docker image, expect key value (could be read from config file)
-
-        :param key: str
-        :param value: str
-        :return: bool
-        """
-        if key in self.backend.containerInfo['Labels'] and (
-                    value in self.backend.containerInfo['Labels'][key]):
-            return True
-        return False
-
-
--- a/moduleframework/helpers/openshift_helper.py
+++ b/moduleframework/helpers/openshift_helper.py
@@ -22,6 +22,7 @@
 
 import json
 import os
+import time
 from moduleframework import common
 from moduleframework.common import CommonFunctions
 from moduleframework.mtfexceptions import ContainerExc, ConfigExc
@@ -52,10 +53,7 @@ class OpenShiftHelper(CommonFunctions):
         # application name is taken from docker.io/modularitycontainer/memcached
         self.app_name = self.container_name.split('/')[-1]
         self.app_ip = None
-
-        common.print_info(self.icontainer)
-        common.print_info(self.container_name)
-        common.print_info(self.app_name)
+        common.print_debug(self.icontainer, self.app_name)
 
     def get_container_url(self):
         """
@@ -78,16 +76,13 @@ class OpenShiftHelper(CommonFunctions):
         :return: True, application exists
                  False, application does not exist
         """
-        common.print_info("OpenShift app_exists")
         oc_status = self.runHost("oc status", ignore_status=True)
         if 'dc/%s' % self.app_name in oc_status.stdout:
             common.print_info("Application already exists.")
             return True
         oc_services = self.runHost("oc get services -o json")
         json_svc = self._convert_string_to_json(oc_services.stdout)
-        common.print_info(json_svc["items"])
         if not json_svc["items"]:
-            common.print_info("itesms are empty")
             return False
         return True
 
@@ -97,9 +92,7 @@ class OpenShiftHelper(CommonFunctions):
         :param string: String to format to json
         :return: json output
         """
-        json_output = json.loads(string)
-        common.print_info(json_output)
-        return json_output
+        return json.loads(string)
 
     def _remove_apps_from_openshift_namespaces(self, oc_service="svc"):
         """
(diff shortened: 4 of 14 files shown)


moduleframework/avocado_testers/openshift_avocado_test.py
```diff
@@ -0,0 +1,54 @@
...
+#
+
+from moduleframework.module_framework import AvocadoTest
+from moduleframework.common import get_module_type_base
+
+
+# INTERFACE CLASSES FOR SPECIFIC MODULE TESTS
+class OpenShiftAvocadoTest(AvocadoTest):
+    """
+    Class for writing tests specific just for DOCKER
```
TomasTomecek: s/DOCKER/OpenShift/

moduleframework/avocado_testers/openshift_avocado_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+    :avocado: disable
+    """
+
+    def setUp(self):
+        if get_module_type_base() != "openshift":
+            self.cancel("OpenShift specific test")
+        super(OpenShiftAvocadoTest, self).setUp()
+
+    def checkLabel(self, key, value):
```
TomasTomecek: is this method up to date?

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,239 @@
...
+        :return: None
+        """
+        common.print_info("OpenShift start")
+        if not self._app_exists():
+            self._create_app()
+            self._get_ip_instance()
+
+    def stop(self):
+        """
+        Stop the docker container
```
TomasTomecek: docstring not up to date

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,239 @@
...
+            return True
+        except KeyError as e:
+            common.print_info(e.message)
+            return False
+
+    def start(self):
+        """
+        start the OpenShift application
+
+        :param args: Do not use it directly (It is defined in config.yaml)
```
TomasTomecek: this method doesn't accept any arguments

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,239 @@
...
+        oc_get_service = self.runHost("oc get service -o json")
+        service = self._convert_string_to_json(oc_get_service.stdout)
+        try:
+            common.print_info(service["items"])
+            common.print_info(service["items"][0])
+            common.print_info(service["items"][0]["spec"])
+            self.app_ip = service["items"][0]["spec"]["clusterIP"]
+            common.trans_dict['GUESTIPADDR'] = self.app_ip
+            common.print_info(common.trans_dict)
+            return True
```
TomasTomecek: docstring says that this method should return an ip address

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,239 @@
...
+        except KeyError as e:
```
TomasTomecek: can also throw IndexError and TypeError (`None[0]`)

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,239 @@
...
+    def _get_ip_instance(self):
+        """
+        Function return IP address of OpenShift POD.
+        :return:
+        """
+        oc_get_service = self.runHost("oc get service -o json")
+        service = self._convert_string_to_json(oc_get_service.stdout)
+        try:
+            common.print_info(service["items"])
+            common.print_info(service["items"][0])
```
TomasTomecek: what if there are other services running?

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def __load_inspect_json(self):
+        """
+        Load json data from docker inspect command
+
+        :return: dict
+        """
+        return json.loads(
+            self.runHost(
+                "docker inspect %s" %
```
TomasTomecek: openshift is not tied to docker

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,239 @@
...
+        * run and store identification
+
+        :return: None
+        """
+        self.icontainer = self.get_url()
+        self.containerInfo = self.__load_inspect_json()
+
+    def _openshift_login(self, oc_ip="127.0.0.1", oc_user='developer', oc_passwd='developer', env=False):
+        if env:
+            if 'OPENSHIFT_IP' in os.environ:
```
TomasTomecek: documentation would be lovely

docs/user_guide/environment_setup.rst
```diff
@@ -21,13 +21,18 @@ Manual Setup
 
  - No any configuration needed
 
+**OpenShift**
+
+ - Install OpenShift if not installed and if environment variable ``OPENSHIFT_LOCAL`` is specified.
+ - if ``OPENSHIFT_LOCAL`` variable is specified, the it starts an OpenShift by command ``oc cluster up`` or stops it by command ``oc cluster down``.
```
TomasTomecek: then*

docs/user_guide/environment_variables.rst
```diff
@@ -25,6 +25,10 @@ Environment variables allow to overwrite some values of a module configuration f
 - **MTF_REMOTE_REPOS=yes** disables downloading of Koji packages and creating a local repo, and speeds up test execution.
 - **MTF_DISABLE_MODULE=yes** disables module handling to use nonmodular test mode (see `multihost tests`_ as an example).
 - **DOCKERFILE="<path_to_dockerfile"** overwrites the location of a Dockerfile.
+- **OPENSHIFT_LOCAL** enables installing ``origin`` and ``origin-clients`` on local machine
+- **OPENSHIFT_IP** uses this IP address for connecting to an OpenShift environment.
+- **OPENSHIFT_USER** uses this ``USER`` name for login to an OpenShift environment.
+- **OPENSHIFT_PWD** uses this ``PWD`` name for login to an OpenShift environment.
```
TomasTomecek: PWD in our world means "print working directory"

docs/user_guide/environment_variables.rst
```diff
@@ -25,6 +25,10 @@ Environment variables allow to overwrite some values of a module configuration f
 - **MTF_REMOTE_REPOS=yes** disables downloading of Koji packages and creating a local repo, and speeds up test execution.
 - **MTF_DISABLE_MODULE=yes** disables module handling to use nonmodular test mode (see `multihost tests`_ as an example).
 - **DOCKERFILE="<path_to_dockerfile"** overwrites the location of a Dockerfile.
+- **OPENSHIFT_LOCAL** enables installing ``origin`` and ``origin-clients`` on local machine
```
TomasTomecek: would be nice to follow the previous definitions: e.g. what is the sample value for this variable?

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,245 @@
...
+#
+
+import json
+import os
+from moduleframework import common
+from moduleframework.common import CommonFunctions
+from moduleframework.mtfexceptions import ContainerExc, ConfigExc
+
+
+class OpenShiftHelper(CommonFunctions):
```
TomasTomecek: So how do I actually use this?

phracek: It is used from commandline `sudo MODULE=openshift mtf` which executes this class.

TomasTomecek: sweet! include it in docs please

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+            if 'OPENSHIFT_USER' in os.environ:
+                oc_user = os.environ.get('OPENSHIFT_USER')
+            if 'OPENSHIFT_PASSWORD' in os.environ:
+                oc_passwd = os.environ.get('OPENSHIFT_PASSWORD')
+
+        oc_output = self.runHost("oc login %s:8443 --username=%s --password=%s" % (oc_ip,
+                                                                                   oc_user,
+                                                                                   oc_passwd),
+                                 verbose=common.is_not_silent())
+        common.print_debug(oc_output.stdout)
```
jscotka: why you use this debugging? isn't better to use `verbose=variable` directly. it exactly does what you want
 `verbose=is_debug()` print via avocado run command output to logger and it is there. in case DEBUG is not set, it is not printed out. (when you don't use it with avocado runner, log is not shown, but it can be enabled.)

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+        It gets and IP address of an application from of OpenShift POD.
+        :return: True: getting IP address was successful
+                 False: getting IP address was not successful
+        """
+        oc_get_service = self.runHost("oc get service -o json")
+        service = self._convert_string_to_json(oc_get_service.stdout)
+        try:
+            for svc in service["items"]:
+                if "clusterIP" in svc.get("spec"):
+                    common.trans_dict['GUESTIPADDR'] = svc.get("spec").get("clusterIP")
```
jscotka: maybe this is proper place where to modify also `self.ipaddr` variable and then you can remove some special handling of ipadd in common.CommonFunciton for OS

and maybe question, why there is __for__ cycle. means that there can be more ipadresses and and the last one is the best one?
or you meant code like:
```
for svc un service.get("items", []):
    ipaddr = svc.get("spec",{}).get("clusterIP")
    if ipaddr:
          common.trans_dict['GUESTIPADDR'] = ipaddr
          self.ipaddr = ipaddr
          return True
return False
```
?

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+            # TODO get info from oc status and delete relevat svc/dc/is
+            for ns in ['svc', 'dc', 'is']:
+                self._remove_apps_from_openshift_namespaces(ns)
+
+    def _create_app(self):
+        """
+        It creates an application in OpenShift environment
+        """
+        # Switching to system user
+        oc_new_app = self.runHost("oc new-app %s --name=%s" % (self.container_name,
```
TomasTomecek: I think you should add a label here to mark the deployment being owned by mtf: `oc new-app -l mtf_testing=true` or something like that; and then operate on objects which have this label.

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def _verify_pod(self):
+        """
+        It verifies if an application POD is initiated and ready for testing
+        :return: False, application is not initiated during 10 seconds
+                 True, application is initiated and ready for testing
+        """
+        pod_initiated = False
+        for x in range(0, 20):
+            pod_state = self.runHost("oc get pods", ignore_status=True)
```
TomasTomecek: 1. please use json and don't parse CLI output

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        self.icontainer = self.get_url()
+
+    def _openshift_login(self, oc_ip="127.0.0.1", oc_user='developer', oc_passwd='developer', env=False):
+        """
+        It logins to an OpenShift environment on specific IP and under user and his password.
+        :param oc_ip: an IP where is an OpenShift environment running
+        :param oc_user: an username under which we can login to OpenShift environment
+        :param oc_passwd: a password for specific username
+        :param env:
```
TomasTomecek: yes?

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+        :return:
```
TomasTomecek: yes?

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        if env:
```
TomasTomecek: I probably don't understand this logic

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def tearDown(self):
+        """
+        Cleanup environment and call also cleanup from config
+
+        :return: None
+        """
+        super(OpenShiftHelper, self).tearDown()
+        if common.get_if_do_cleanup():
+            # TODO will be implemented later on. I have to find a usecase what to remove
```
TomasTomecek: I would probably remove everything which was put into openshift during this test

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+        if self.status():
+            try:
+                self._app_remove()
+            except Exception as e:
+                common.print_info(e, "OpenShift application already removed")
+                pass
+
+    def status(self):
+        """
+        get status of an OpenShift
```
TomasTomecek: doesn't seem to correspond with the actual code

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+
+        :return: bool
+        """
+        if self._app_exists():
```
TomasTomecek: you can easily `return self._app_exists()`

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def run(self, command="ls /", **kwargs):
+        """
+        Run command inside module, all params what allows avocado are passed inside shell,ignore_status, etc.
+
+        :param command: str
+        :param kwargs: dict
+        :return: avocado.process.run
+        """
+        return self.runHost("%s" % common.sanitize_cmd(command), **kwargs)
```
jscotka: Are you sure that this does what you expect?
this run command on host.
`run` method of helper should do command on __GUEST__ in this case on openshift instance in case of docker there is for example `docker exec INSTANCE_ID command` so in case of openshift I expect something similar like `oc exec ...` 

it is connected with more missing special methods for openshift, like `CopyFrom` and `CopyTo` now are inherited from CommonFunctions, what leads to `host to host` implementation, I expect there also some special `oc ` command, or do it via some ssh, or whatever how it is connected together

Probably my mistake were that I've defined these functions leads to `localhost` operations, instead of raising not defined exceptions, and implement these methods where it make sense like in `RpmHelper, ContainerHelper` etc...

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,268 @@
...
+        :return: True, application exists
+                 False, application does not exist
+        """
+        oc_status = self.runHost("oc status", ignore_status=True)
+        if 'dc/%s' % self.app_name in oc_status.stdout:
+            common.print_info("Application already exists.")
+            return True
+        oc_services = self.runHost("oc get services -o json")
+        json_svc = self._convert_string_to_json(oc_services.stdout)
+        if not json_svc["items"]:
```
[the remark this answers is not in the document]
jscotka: as @TomasTomecek mentioned, probably you need to add there some "__service__" specific calls. I don't know it but probably there will be something similar to `docker container inspect CONTAINER_ID`  what will show state just for one service instead of list all services and try to get some data

moduleframework/helpers/openshift_helper.py
jscotka: @phracek @TomasTomecek yep, similar to `docker container ...` commands and then you will just use this tag. otherwise I expect that this __iddentifier__ will be stored in OpenShiftHelper class like `self.jmeno` for `ContainerHelper`

moduleframework/helpers/openshift_helper.py
phracek: Does this make sense to you?
```env: is used for specification OpenShift IP, user and password, otherwise defaults are used```

File: moduleframework/helpers/openshift_helper.py (hunk 19 above)
Comment (phracek) [reply moduleframework/helpers/openshift_helper.py]: Comment should be: ```env: is used for specification OpenShift IP, user and password, otherwise defaults are used```

File: moduleframework/helpers/openshift_helper.py [hunk 26]
```diff
@@ -0,0 +1,268 @@
...
+        It removes an application from specific "namespace" like svc, dc, is.
+        :param oc_service: Service from which we would like to remove application
+        """
+        # Check status of svc/dc/is
+        oc_get = self.runHost("oc get %s" % oc_service, ignore_status=True).stdout
+        # The output is like
+        # dovecot     172.30.1.1:5000/myproject/dovecot     latest    15 minutes ago
+        # memcached   172.30.1.1:5000/myproject/memcached   latest    13 minutes ago
+
+        app_found = [x for x in oc_get.split('\n') if x.startswith(self.app_name)]
```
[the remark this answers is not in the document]
phracek: I will try to use it.

moduleframework/helpers/openshift_helper.py
TomasTomecek: no, I still don't understand the reason for that variablee, since the values are being picked up from environment

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,300 @@
...
+        if self.status():
+            try:
+                self._app_remove()
+            except Exception as e:
+                common.print_info(e, "OpenShift application already removed")
+                pass
+
+    def status(self):
+        """
+        get status of an application in OpenShift environment
```
TomasTomecek: I still think that this docstring doesn't correspond with the code: what does the "status" mean exactly? For me, status is usually enum, something like "running", "failed", "finished" and this method returns bool. It makes me confused.

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+        if not self._app_exists():
+            self._create_app()
+            # Verify application is really deploy and prepared for testing.
+            self._verify_pod()
+        self._get_ip_instance()
+
+    def stop(self):
+        """
+        Stops the OpenShift
```
TomasTomecek: another misleading docstring; I think this method actually removes the application from the cluster

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,300 @@
...
+
+        :return: str
+        """
+        return self.ipaddr
+
+    def start(self):
+        """
+        starts the OpenShift application
+
+        :param args: Do not use it directly (It is defined in config.yaml)
```
TomasTomecek: this function doesn't accept any args

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,300 @@
...
+                    common.trans_dict['GUESTIPADDR'] = self.ipaddr
+            return True
+        except KeyError as e:
+            common.print_info(e.message)
+            return False
+        except IndexError as e:
+            common.print_info(e.message)
+            return False
+
+    def getIPaddr(self):
```
TomasTomecek: pythonic way of doing setters and getters is like this: https://docs.python.org/3/library/functions.html#property

(not blocking this review, just a friendly advice)

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+        It gets and IP address of an application from of OpenShift POD.
+        :return: True: getting IP address was successful
+                 False: getting IP address was not successful
+        """
+        oc_get_service = self.runHost("oc get service -o json")
+        service = self._convert_string_to_json(oc_get_service.stdout)
+        try:
+            for svc in service:
+                if svc.get('metadata').get('labels').get('app') == self.app_name:
```
TomasTomecek: +1

moduleframework/helpers/openshift_helper.py
```diff
@@ -0,0 +1,300 @@
...
+        super(OpenShiftHelper, self).tearDown()
+        try:
+            self._app_remove()
+        except Exception as e:
+            common.print_info(e, "OpenShift application already removed")
+            pass
+
+    def _get_ip_instance(self):
+        """
+        It gets and IP address of an application from of OpenShift POD.
```
TomasTomecek: How about: "This method verifies that we can obtain an IP address of the application deployed within OpenShift."

moduleframework/helpers/openshift_helper.py
phracek: https://github.com/fedora-modularity/meta-test-family/issues/167

moduleframework/helpers/openshift_helper.py
phracek: This should be solved in whole MTF. `status` function always return `bool` value like e.g. here: https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/common.py#L552
https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/helpers/container_helper.py#L171

I have filed an issue for it here: https://github.com/fedora-modularity/meta-test-family/issues/168

I will keep it now as it is.

phracek:
Hopefully, I have addressed all issues. The rest are filed as a new issues.

PR #169, opened 2017-11-16
add some tags to modulelint tests, to be able to filter them

jscotka:
if you would like to use it, you can filter it now via avocado tag filters

### DIFF
--- a/moduleframework/tools/__init__.py
+++ b/moduleframework/tools/__init__.py
@@ -28,6 +28,7 @@ from moduleframework import module_framework
 class Basic(module_framework.AvocadoTest):
     """
     :avocado: enable
+    :avocado: tags=sanity,general,fedora,rhel
     """
 
     def test(self):
--- a/moduleframework/tools/check_compose.py
+++ b/moduleframework/tools/check_compose.py
@@ -33,6 +33,7 @@ class ComposeTest(module_framework.NspawnAvocadoTest):
     Validate overall module compose.
 
     :avocado: enable
+    :avocado: tags=sanity,rhel,fedora,compose_test,module
     """
 
     def test_component_profile_installability(self):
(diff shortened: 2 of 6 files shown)


phracek:
@jscotka LGTM. Is it still valid? Based on the PR with metadata?

phracek:
Nevertheless, where are defined, or even documented tags?

jscotka:
@phracek  yes this is still valid and not anyhow dependent on PR with metadata tooling (even though this parser will use it somehow to demonstrate principles)

phracek:
@jscotka Does it make sense to dokument it?

moduleframework/tools/__init__.py
```diff
@@ -28,6 +28,7 @@
 class Basic(module_framework.AvocadoTest):
     """
     :avocado: enable
+    :avocado: tags=sanity,general,fedora,rhel
```
phracek: I would prefer to document it, so maintainer can create tests for relevant tags. Or at least some documentation how to create and use new tags.

jscotka: It is expected that maintainer will use own tags as @bkabrda mentioned, that he is interested in own tags, like add tag like `network` to some own, to indicate and filter out that some test needs for example network available.  
These tags are now just for modulelint, and probably just: `sanity` and `optional` will be used for now. Other tags will be used more less just an example of more complex filtering abilities rather than officially document list of them somewhere.

And you are talking about how to create and use new tags, and this is avocado specific and documented well, and it is not anyhow connected with MTF: 
http://avocado-framework.readthedocs.io/en/56.0/WritingTests.html#categorizing-tests
If you will use mtf as backend, for example via behave, or unittest, it will not be anyhow used, so this help and link is not generally applicable for MTF.

PR #172, opened 2017-11-21
WIP: added example of how to work with ODCS

jscotka:
Future plan is to have there just __ODCS__ support and remove koji downloading and similar hacks and workaround.
https://pagure.io/odcs/
I plan to work it on that extensively,  but comments are welcomed


moduleframework/common.py
```diff
@@ -242,6 +242,10 @@ def get_if_remoterepos():
     remote_repos = os.environ.get('MTF_REMOTE_REPOS')
     return bool(remote_repos)
 
+def get_odcs_auth():
+    odcstoken = os.environ.get('MTF_ODCS')
```
TomasTomecek: would be awesome if ODCS supported this on its own

moduleframework/pdc_data.py
```diff
@@ -299,13 +300,24 @@ def createLocalRepoFromKoji(self):
...
+    def get_module_identifier(self):
+        if self.version:
+            return "%s-%s-%s" % (self.name, self.stream, self.version)
+        elif self.stream:
+            return "%s-%s" % (self.name, self.stream)
+        else:
+            return "%s-%s" % (self.name, "master")
+
+    def get_odcs_compose_repo(self, auth_token, compose_type="module"):
+        odcs = ODCS(ODCS_URL, auth_mech=AuthMech.OpenIDC, openidc_token=auth_token)
```
TomasTomecek: Would probably deserve a dedicated class on its own.

moduleframework/common.py
jscotka: It would be nice. we ca file issue and in future, try to change it.

moduleframework/pdc_data.py
jscotka: Sounds well, probably we need to rewrite this pdc_data.py completely, actually there are mixed koji, modulemd, pdc, odcs handling. 
It is caused by that complete module info is not in PDC, but composed by using more resources, pdc and koji and modulemds.
I can imagine that we'll have there some generic class for pdc, what will be extended by `koji`, second class extended by `odcs`. and then choose proper class based on `envvars`, does that sense @TomasTomecek

TomasTomecek: yup, makes sense to separate them

jscotka:
moved to new PR 
#181

PR #175, opened 2017-11-22
WIP: man mtf page is generated

psklenar:


praiskup:
FWIW, I would drop the generated man page from git.  Nit: it seems that you have misconfigured IDE so it doesn't put newlines at the end of python files.

psklenar:
@praiskup mtf.1 from git was deleted
thanks for the info

moduleframework/mtf_scheduler.py
```diff
@@ -33,24 +33,51 @@
...
+
+       MODULE defines tested module type, if default-module is not set in config.yaml.
+
+       URL to container.
+              E.g. URL=docker.io/modularitycontainers/haproxy if MODULE=docker
+
+       MODULEMDURL
+              overwrites the location of a moduleMD file.
+
+       and more at http://meta-test-family.readthedocs.io/en/latest/user_guide/environment_variables.html
```
phracek: I would prefer to mention them.
`man` command is faster then looking at some web page.

setup.cfg
```diff
@@ -1,2 +1,6 @@
 [metadata]
 description-file = README.md
+
+[build_manpages]
+manpages =
+    man/mtf.1:function=mtfparser:module=moduleframework.mtf_scheduler
```
phracek: I guess, we have already several man pages. Do you plan to mention them?

psklenar: right, they can be generated too. I will create new issue about that.

moduleframework/mtf_scheduler.py
psklenar: are we sure about that? when I saw that, man page seems too long to me, but I am not typical "manpage reader"

PR #179, opened 2017-11-28
Propery setter getter

phracek:
This Pull Request implements one feature. @property, setter and getter.

I have tried to find out all issues, where we can use it and hopefully, I catched all of them.

### DIFF
--- a/moduleframework/avocado_testers/avocado_test.py
+++ b/moduleframework/avocado_testers/avocado_test.py
@@ -26,9 +26,9 @@ main module provides all helpers for various module types and AVOCADO(unittest) 
 what you should use for your tests (inherited)
 """
 
-
 from avocado import Test
 from avocado.core import exceptions
+import warnings
 
 from moduleframework.common import *
 from moduleframework.helpers.container_helper import ContainerHelper
@@ -226,6 +226,17 @@ class AvocadoTest(Test):
         """
         return self.backend.copyFrom(*args, **kwargs)
 
+    def getIPaddr(self, *args, **kwargs):
+        """
+        Return ip addr string of guest machine
+        In many cases it should be same as host machine and port should be forwarded to host
+
+        :return: str
+        """
+        warnings.warn("Function getIPaddr is deprecated. Use self.ipaddr instead",
+                      DeprecationWarning)
+        return self.backend.ipaddr(*args, **kwargs)
+
     def ipaddr(self, *args, **kwargs):
         """
         Return ip addr string of guest machine
@@ -242,6 +253,18 @@ class AvocadoTest(Test):
         :return: str
         """
         return self.backend.getArch()
+
+    def getModuleDependencies(self):
+        """
+        get list of module dependencies dictionary, there is structure like:
+        {module_name: {stream: master, urls=[repo_url1, repo_url2]},
+         dependent_module_name: {stream: f26, urls=[repo_url3]}}
+
+        :return: dict
+        """
+        warnings.warn("Function getModuleDependencies is deprecated. Use self.dependency_list instead",
+                      DeprecationWarning)
+        return self.dependency_list()
 
     def dependency_list(self):
         """
--- a/moduleframework/avocado_testers/avocado_test.py
+++ b/moduleframework/avocado_testers/avocado_test.py
@@ -228,23 +228,24 @@ class AvocadoTest(Test):
 
     def getIPaddr(self, *args, **kwargs):
         """
-        Return ip addr string of guest machine
+        Return ip address string of guest machine
         In many cases it should be same as host machine and port should be forwarded to host
 
         :return: str
         """
-        warnings.warn("Function getIPaddr is deprecated. Use self.ipaddr instead",
+        warnings.warn("Function getIPaddr is deprecated. Use self.ip_address instead",
                       DeprecationWarning)
-        return self.backend.ipaddr(*args, **kwargs)
+        return self.backend.ip_address(*args, **kwargs)
 
-    def ipaddr(self, *args, **kwargs):
+    @property
+    def ip_address(self, *args, **kwargs):
         """
-        Return ip addr string of guest machine
+        Return ip address string of guest machine
         In many cases it should be same as host machine and port should be forwarded to host
 
         :return: str
         """
-        return self.backend.ipaddr(*args, **kwargs)
+        return self.backend.ip_address(*args, **kwargs)
 
     def getArch(self):
         """
@@ -266,6 +267,7 @@ class AvocadoTest(Test):
                       DeprecationWarning)
         return self.dependency_list()
 
+    @property
     def dependency_list(self):
         """
         get list of module dependencies dictionary, there is structure like:
--- a/moduleframework/avocado_testers/avocado_test.py
+++ b/moduleframework/avocado_testers/avocado_test.py
@@ -226,14 +226,14 @@ class AvocadoTest(Test):
         """
         return self.backend.copyFrom(*args, **kwargs)
 
-    def getIPaddr(self, *args, **kwargs):
+    def ipaddr(self, *args, **kwargs):
         """
         Return ip addr string of guest machine
         In many cases it should be same as host machine and port should be forwarded to host
 
         :return: str
         """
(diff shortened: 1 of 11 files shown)


phracek:
@jscotka @TomasTomecek @bkabrda Could you please guys, someone, review this Pull Request?

moduleframework/helpers/container_helper.py
```diff
@@ -37,41 +37,50 @@ def __init__(self):
...
         :return: str
         """
-        return self.icontainer
+        return self._icontainer
+
+    @URL.setter
+    def URL(self, value):
+        self._icontainer = value
+
+    @URL.getter
```
TomasTomecek: you already define
```python
@property
def URL(...):
```

so you can remove this explicit getter

moduleframework/common.py
```diff
@@ -499,14 +501,23 @@ def getPackageList(self, profile=None):
         print_info("PCKGs to install inside module:", package_list)
         return package_list
 
-    def getModuleDependencies(self):
+    @property
+    def dependencylist(self):
```
TomasTomecek: doesn't comply to [PEP8](https://www.python.org/dev/peps/pep-0008/#function-names):

    Function names should be lowercase, with words separated by underscores as necessary to improve readability.

moduleframework/helpers/container_helper.py
```diff
@@ -37,41 +37,50 @@ def __init__(self):
         set basic object variables
         """
         super(ContainerHelper, self).__init__()
-        static_name="testcontainer"
+        static_name = "testcontainer"
         self.tarbased = None
-        self.jmeno = None
+        self.name = None
```
TomasTomecek: +1

moduleframework/common.py
```diff
@@ -540,7 +551,8 @@ def getModulemdYamlconfig(self, urllink=None):
             self.modulemdConf = link
         return link
 
-    def getIPaddr(self):
+    @property
+    def ipaddr(self):
```
TomasTomecek: You are probably changing your public API here, is that correct?

phracek: Yes, Previously, code looks like `self.getIPaddr()` which returns proper IP and now is easy to use `self.ipaddr`.

@jscotka Your opinion is also welcome. 

Does it make sense?

moduleframework/helpers/container_helper.py
phracek: I see, I though that `getter` should be defined always. THX.

moduleframework/common.py
phracek: @jscotka I would like to test if dependency_list works properly. Any idea? Do we have a test for it already in `examples` directory?

moduleframework/common.py
TomasTomecek: My point is: if anyone is using `self.getIPaddr()` in their tests, this change will break that test and people will need to fix their tests, are you okay with that?

TomasTomecek:
(needs rebase)

moduleframework/common.py
jscotka: @phracek exactly, we have to be backward compatible.
so yep, do that via these getter/setter but have there these old methods too (what will be just wrappers to these new getters/setters)

TomasTomecek: usually projects utilize [`warnings`](https://docs.python.org/2/library/warnings.html) module for such changes and tell people that such functionality is deprecated and will be removed in release X

moduleframework/helpers/container_helper.py
jscotka: @TomasTomecek @phracek I'm not against. but I would like to change it to something different, like
`self.some_own_identifier`. I've choosed `self.jmeno` before because it is not name and should not lead to `name` interpretation, it is something what is close to `name` but this is not name, but more less some more less unique identifier for various artifacts.

moduleframework/common.py
```diff
@@ -580,7 +586,23 @@ def getIPaddr(self):
...
+        Return protocol (IP or IPv6) address on a guest machine.
+
+        In many cases it should be same as a host machine's and a port
+        should be forwarded to a host.
+
+        :return: str
+        """
+        return self._ipaddr
+
+    @ipaddr.setter
```
jscotka: I'm not sure with this, do you think that ipaddr should be changed from outside? it should be set when artifacts starts, after network is initialized. not from outside.

moduleframework/avocado_testers/avocado_test.py
```diff
@@ -243,15 +243,15 @@ def getArch(self):
         """
         return self.backend.getArch()
 
-    def getModuleDependencies(self):
+    def dependency_list(self):
```
jscotka: please let there backward compatibility for every function. Probably as @TomasTomecek with using warning.warn(...) that it will be removed in future, to give chance to users to change behaviour.

moduleframework/helpers/container_helper.py
```diff
@@ -37,41 +37,46 @@ def __init__(self):
...
         self.docker_id = None
-        self.icontainer = self.get_url()
-        if not self.icontainer:
+        self._icontainer = self.get_url()
```
jscotka: Major part will be replaced by __CONU__ projet, so do not spent effort there.
Also `self.jmeno` will be removed after using CONU.
@TomasTomecek what do you thing. Does that make sense to repair it, or just let it as it is, and then remove all own code by CONU?

moduleframework/helpers/rpm_helper.py
```diff
@@ -46,14 +47,19 @@ def __init__(self):
...
         """
         Return semicolon separated string of repositories what will be used, could be simialr to URL param,
          it contains also dependent repositories from PDC
 
         :return: str
         """
-        return ";".join(self.repos)
+        return ';'.join(self.repos)
+
+    @URL.setter
```
jscotka: this URL, should be just readonly, it should not be  allowed SET URL

phracek: @jscotka And how would you like to setup? in __init__ function?

moduleframework/common.py
phracek: @jscotka `self.ipaddr = <something>` sets the ipaddress. `self._ipaddr` should not be readable directly.

moduleframework/avocado_testers/avocado_test.py
```diff
@@ -233,7 +233,18 @@ def getIPaddr(self, *args, **kwargs):
 
         :return: str
         """
-        return self.backend.getIPaddr(*args, **kwargs)
+        warnings.warn("Function getIPaddr is deprecated. Use self.ipaddr instead",
+                      DeprecationWarning)
+        return self.backend.ipaddr(*args, **kwargs)
+
+    def ipaddr(self, *args, **kwargs):
```
TomasTomecek: should this be wrapped with `@property`?

moduleframework/avocado_testers/avocado_test.py
```diff
@@ -233,7 +233,18 @@ def getIPaddr(self, *args, **kwargs):
...
+        """
+        Return ip addr string of guest machine
```
TomasTomecek: s/addr/address/

moduleframework/avocado_testers/avocado_test.py
TomasTomecek: also the name doesn't comply to pep 8; I would personally name it `ip_address`; the other thing is IPv4 and IPv6

moduleframework/avocado_testers/avocado_test.py
```diff
@@ -251,7 +262,19 @@ def getModuleDependencies(self):
 
         :return: dict
         """
-        return self.backend.getModuleDependencies()
+        warnings.warn("Function getModuleDependencies is deprecated. Use self.dependency_list instead",
+                      DeprecationWarning)
+        return self.dependency_list()
+
+    def dependency_list(self):
```
TomasTomecek: same thing, doesn't this miss `@property`?

moduleframework/avocado_testers/avocado_test.py
phracek: I would not complicated for now. Let's start with ip_address and later on we can enhance it for IPv6.

PR #188, opened 2017-12-04
Metadata support to mtf tool, enabled via option --metadata

jscotka:


jscotka:
maybe you are searching for:  `examples/test_metadata_loader`, it contains PoC how this can be used (also contains `Makefile`) what does more, but also this PoC, how it can be used:
`MODULE=docker URL=fedora mtf --metadata `

examples/test_metadata_loader/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+document: test-metadata
+subtype: general
+enable_lint: True
+tag_filters:
+    - "docker_labels_inspect_test"
+import_tests:
+    - "*.py"
+    - "/bin/true"
```
TomasTomecek: this is pretty interesting test

jscotka: yep, I were lazy to write any useful test, this dummy test is enough to schedule it via avocado :-)

PR #194, opened 2017-12-18
OpenShift fixes catched by real testing

phracek:

This Pull Request brings couple fixes catched by real testing.
The changes affected only OpenShift.

### DIFF
--- a/moduleframework/avocado_testers/avocado_test.py
+++ b/moduleframework/avocado_testers/avocado_test.py
@@ -238,7 +238,7 @@ class AvocadoTest(Test):
         return self.backend.ip_address
 
     @property
-    def ip_address(self, *args, **kwargs):
+    def ip_address(self):
         """
         Return ip address string of guest machine
         In many cases it should be same as host machine and port should be forwarded to host
--- a/moduleframework/helpers/openshift_helper.py
+++ b/moduleframework/helpers/openshift_helper.py
@@ -43,7 +43,7 @@ class OpenShiftHelper(ContainerHelper):
         self.name = None
         self.icontainer = self.get_url()
         self.pod_id = None
-        self._pod_status = None
+        self._ip_address = None
         if not self.icontainer:
             raise ConfigExc("No container image specified in the configuration file or environment variable.")
         if "docker=" in self.icontainer:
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -103,5 +103,5 @@ def get_dir(system_path=None, virtual_path=None):
         'Topic :: Software Development',
     ],
     install_requires=open('requirements.txt').read().splitlines(),
-    zip_safe=True
```
TomasTomecek: I thought it was set to True to fix a packaging issue.

moduleframework/helpers/openshift_helper.py
```diff
@@ -66,7 +66,7 @@ def _app_exists(self):
         if int(oc_status.exit_status) == 0:
             common.print_info("Application already exists.")
             return True
-        oc_services = self.runHost("oc get services -o json", ignore_status=True).stdout
+        oc_services = self.runHost("oc get pods -o json", ignore_status=True).stdout
```
TomasTomecek: this seem odd, are you sure you know what you are doing?

phracek: I will clarify it. `pods` make sense, but what if application failed. And the other use cases.

moduleframework/avocado_testers/avocado_test.py
```diff
@@ -226,7 +226,7 @@ def copyFrom(self, *args, **kwargs):
         """
         return self.backend.copyFrom(*args, **kwargs)
 
-    def getIPaddr(self, *args, **kwargs):
```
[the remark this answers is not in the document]
phracek: I don't see any reason, why to use args or kwargs.

setup.py
phracek: My fault. You are right.

moduleframework/avocado_testers/avocado_test.py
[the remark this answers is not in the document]
jscotka: I mean other methods in this class like: getModuleYaml and more: https://github.com/phracek/meta-test-family/blob/46f542b7c9678544d53dc2bf1d2d9b54966e3093/moduleframework/avocado_testers/avocado_test.py#L188 
almost all methods contais this pass thru arguments and probably would be nice to reconsider it and remove all these param forwarding and have there explicit params (to be easier to use also code completion)

moduleframework/helpers/openshift_helper.py
phracek: Based on discussion with OpenShift folks, `oc get pods` is better for getting status if application is READY.

phracek:
@jscotka Yes, but I have to run `oc cluster up`.  And unfortunally OpenShift server is not on PyPi.
I am trying to solve it by separate PR. https://github.com/fedora-modularity/meta-test-family/pull/195

PR #195, opened 2018-01-05
Include OpenShift into Travis CI and testing

phracek:

This Pull Request introduce testing container inside OpenShift.
As a testing container memcached is used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
   - sudo make -C examples/testing-module check-real-rpm-destructive
   - sudo make -C examples/testing-module check-docker-scl-multi-travis
   #- sudo make -C examples/testing-module check-mtf-pdc-module-info-reader
-  #- sudo make -C mtf/metadata check
-  #- sudo make -C examples/testing-module check-mtf-metadata
+  - sudo make -C mtf/metadata check
+  - sudo make -C examples/testing-module check-mtf-metadata
 
 after_script: sudo cat /home/travis/avocado/job-results/latest/job.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,17 +14,31 @@ before_install:
 
 install: sudo make install_pip
 
-script:
-  - sudo make travis
-  - sudo make -C examples/testing-module check-inheritance
-  - sudo make -C examples/testing-module check-default-config
-  - sudo make -C examples/testing-module check-pure-docker
-  - sudo make -C examples/testing-module check-exceptions
-  - sudo make -C examples/testing-module check-test-mtf-bin-modulelint
-  - sudo make -C examples/testing-module check-real-rpm-destructive
-  - sudo make -C examples/testing-module check-docker-scl-multi-travis
-  - sudo make -C examples/testing-module check-mtf-pdc-module-info-reader
-  - sudo make -C mtf/metadata check
-  - sudo make -C examples/testing-module check-mtf-metadata
+before_script:
+  - |
+    wget https://github.com/openshift/origin/releases/download/v3.6.1/openshift-origin-client-tools-v3.6.1-008f2d5-linux-64bit.tar.gz -O ~/openshift-origin-client-tools-v3.6.1-008f2d5-linux-64bit.tar.gz
+    tar -xzvf ~/openshift-origin-client-tools-v3.6.1-008f2d5-linux-64bit.tar.gz
+    sudo ln -s $PWD/openshift-origin-client-tools-v3.6.1-008f2d5-linux-64bit/oc /usr/bin/oc
+    tmp=`mktemp`
+    echo 'DOCKER_OPTS="$DOCKER_OPTS --insecure-registry 172.30.0.0/16"' > ${tmp}
+    sudo mv ${tmp} /etc/default/docker
+    sudo mount --make-shared /
+    sudo service docker restart
+    docker info
+    export PATH=$PATH:$PWD/openshift-origin-client-tools-v3.7.0-7ed6862-linux-64bit/
 
-after_script: sudo cat ~/avocado/job-results/latest/job.log
+script:
+  - sudo make -C examples/testing-module check-openshift-memcached
+  #- sudo make travis
+  #- sudo make -C examples/testing-module check-inheritance
+  #- sudo make -C examples/testing-module check-default-config
+  #- sudo make -C examples/testing-module check-pure-docker
+  #- sudo make -C examples/testing-module check-exceptions
+  #- sudo make -C examples/testing-module check-test-mtf-bin-modulelint
+  #- sudo make -C examples/testing-module check-real-rpm-destructive
+  #- sudo make -C examples/testing-module check-docker-scl-multi-travis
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -15,7 +15,11 @@ before_install:
 install: sudo make install_pip
 
 script:
+  - sudo wget https://github.com/openshift/origin/releases/download/v3.7.0/openshift-origin-client-tools-v3.7.0-7ed6862-linux-64bit.tar.gz
+  - sudo tar -xzvf ./openshift-origin-client-tools-v3.7.0-7ed6862-linux-64bit.tar.gz
+  - sudo ln -s ./openshift-origin-client-tools-v3.7.0-7ed6862-linux-64bit/oc /home/travis/oc
```
TomasTomecek: why not adding the symlink to $PATH directory?

examples/testing-module/Makefile
```diff
@@ -76,6 +76,10 @@ check-memcached-both: prepare-docker prepare-nspawn
 	cd ../memcached; MODULE=docker $(CMD) sanity1.py
 	cd ../memcached; MODULE=nspawn $(CMD) sanity1.py
 
+check-openshift-memcached:
+	sudo /home/travis/oc cluster up
```
TomasTomecek: ...because this won't work locally

.travis.yml
phracek: thx, I did it, but `oc` is not visible for some reason.

```
$ sudo make -C examples/testing-module check-openshift-memcached
make: Entering directory `/home/travis/build/fedora-modularity/meta-test-family/examples/testing-module'
sudo oc cluster up
sudo: oc: command not found
make: *** [check-openshift-memcached] Error 1
make: Leaving directory `/home/travis/build/fedora-modularity/meta-test-family/examples/testing-module'
```

phracek:
OpenShift is integrated, but failed on metadata or pdc-metadata. Test are unstable.

.travis.yml
```diff
@@ -14,17 +14,31 @@ before_install:
...
   - sudo make -C examples/testing-module check-default-config
   - sudo make -C examples/testing-module check-pure-docker
   - sudo make -C examples/testing-module check-exceptions
   - sudo make -C examples/testing-module check-test-mtf-bin-modulelint
   - sudo make -C examples/testing-module check-real-rpm-destructive
   - sudo make -C examples/testing-module check-docker-scl-multi-travis
-  - sudo make -C examples/testing-module check-mtf-pdc-module-info-reader
-  - sudo make -C mtf/metadata check
-  - sudo make -C examples/testing-module check-mtf-metadata
+  #- sudo make -C examples/testing-module check-mtf-pdc-module-info-reader
```
jscotka: This case could happen, that it starts failing, because therer were bigger chanes in modularity, and some data was removed. Module data were removed from PDC previous week, so it may affect this

.travis.yml
```diff
@@ -14,17 +14,31 @@ before_install:
...
+  #- sudo make -C mtf/metadata check
```
jscotka: **please do not remove these two tests for metadata**,
 they works after my fixes . So in case this PR https://github.com/fedora-modularity/meta-test-family/pull/196, will be based on latest devel, it should work well.

examples/openshift-sample/config.yaml
```diff
@@ -0,0 +1,9 @@
+document: meta-test
+version: 1
+name: memcached
+service:
+    port: 11211
+module:
+    openshift:
```
jscotka: why isn't this part of ``examples/memcached`` and is there created new directorory for same component? Is that intentional? and is there some big differences. or this mean that ``examples/memcached/sanity1.py`` test does not work with openshift, so that just some simple one works?

examples/testing-module/Makefile
```diff
@@ -76,6 +76,10 @@ check-memcached-both: prepare-docker prepare-nspawn
 	cd ../memcached; MODULE=docker $(CMD) sanity1.py
 	cd ../memcached; MODULE=nspawn $(CMD) sanity1.py
 
+check-openshift-memcached:
+	sudo oc cluster up
```
jscotka: this is not part of ``mtf-env-set`` tool? for **openshift** support (same for down command)

moduleframework/helpers/openshift_helper.py
```diff
@@ -173,10 +176,11 @@ def _verify_pod(self):
                  True, application is initiated and ready for testing
         """
         pod_initiated = False
-        for x in range(0, 20):
+        for x in range(0, 50):
```
jscotka: it is possible to have there some predefined constant, defined for example in common.py like:
``OPENSHIFT_INIT_WAIT = 50`` and then use it there.
Just fomal note. see example like: 
https://github.com/fedora-modularity/meta-test-family/blob/devel/moduleframework/common.py#L89

.travis.yml
```diff
@@ -14,17 +14,31 @@ before_install:
...
+    tar -xzvf ~/openshift-origin-client-tools-v3.6.0-c4dd4cf-linux-64bit.tar.gz
+    sudo ln -s $PWD/openshift-origin-client-tools-v3.6.0-c4dd4cf-linux-64bit/oc /usr/bin/oc
+    tmp=`mktemp`
+    echo 'DOCKER_OPTS="$DOCKER_OPTS --insecure-registry 172.30.0.0/16"' > ${tmp}
+    sudo mv ${tmp} /etc/default/docker
+    sudo mount --make-shared /
+    sudo service docker restart
+    docker info
+    export PATH=$PATH:$PWD/openshift-origin-client-tools-v3.6.0-c4dd4cf-linux-64bit/
+
```
jscotka: why this setup? ``mtf-env-set`` is not able to do these steps, or this is caused by version of ``docker`` on travis?

phracek: OpenShift is not on PyPi. I have to download it, untar it, make some steps. I guess, this is not part for `mtf-env-set`.

examples/testing-module/Makefile
phracek: By the way this is kind of problematic. `mtf-env-set` installs some packages from Fedora repository. And OpenShift is not available in Ubuntu.

examples/openshift-sample/config.yaml
phracek: Problem is that memcached listens on port which is already used by Travis CI. Especially port 11211.

.travis.yml
phracek: In my last Travis Build it fails:(

.travis.yml
jscotka: yep, make sense thanks for explanation

examples/testing-module/Makefile
jscotka: yep, make sense

examples/openshift-sample/config.yaml
jscotka: yes, that's a problem. but still possible to integrate it, and just run some simple test instead of this what needs port mapping. but okay, we can have it in that way

PR #197, opened 2018-01-09
remove avocado.run and replace it by subprocess to see outputs on the fly

jscotka:


moduleframework/mtf_scheduler.py
```diff
@@ -227,8 +227,8 @@ def avocado_run(self):
             ACTION=self.args.action, AVOCADO_ARGS=avocado_args)
 
         # run avocado with right cmd arguments
-        bash = process.run("{AVOCADO} {a} {b}".format(
-            AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=" ".join(self.tests)), shell=True, ignore_status=True)
+        bash = subprocess.check_call("{AVOCADO} {a} {b}".format(
+            AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=" ".join(self.tests)), shell=True)
         common.print_info(bash.stdout, bash.stderr)
```
phracek: This line has to be removed.

moduleframework/mtf_scheduler.py
```diff
@@ -227,8 +227,8 @@ def avocado_run(self):
...
         common.print_debug("Command used: ", bash.command)
```
phracek: As well as this.

moduleframework/mtf_scheduler.py
```diff
@@ -227,8 +227,8 @@ def avocado_run(self):
...
         return bash.exit_status
```
phracek: I would prefer to use exit_code from subprocess. https://docs.python.org/2/library/subprocess.html
And use Popen which returns return code. https://docs.python.org/2/library/subprocess.html#subprocess.Popen.returncode

moduleframework/mtf_scheduler.py
```diff
@@ -239,8 +239,8 @@ def avocado_general(self):
         avocado_args = ""  # when needed => specify HERE your additional stuff
         avocadoAction = "avocado {ACTION} {AVOCADO_ARGS}".format(
             ACTION=self.args.action, AVOCADO_ARGS=avocado_args)
-        bash = process.run("{AVOCADO} {a} {b}".format(
-            AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=" ".join(self.tests)), shell=True, ignore_status=True)
+        bash = subprocess.check_call("{AVOCADO} {a} {b}".format(
+            AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=" ".join(self.tests)), shell=True)
         common.print_info(bash.stdout, bash.stderr)
```
phracek: This has to be removed.

moduleframework/mtf_scheduler.py
```diff
@@ -239,8 +239,8 @@ def avocado_general(self):
         avocado_args = ""  # when needed => specify HERE your additional stuff
         avocadoAction = "avocado {ACTION} {AVOCADO_ARGS}".format(
             ACTION=self.args.action, AVOCADO_ARGS=avocado_args)
-        bash = process.run("{AVOCADO} {a} {b}".format(
-            AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=" ".join(self.tests)), shell=True, ignore_status=True)
+        bash = subprocess.check_call("{AVOCADO} {a} {b}".format(
```
phracek: Use try and expect CalledProcessError.https://docs.python.org/2/library/subprocess.html#subprocess.check_call

moduleframework/mtf_scheduler.py
```diff
@@ -188,24 +189,26 @@ def cli():
...
 
         for param in unknown:
             # take care of this, see tags for safe/unsafe:
             # http://avocado-framework.readthedocs.io/en/52.0/WritingTests.html#categorizing-tests
-            if os.path.exists(param):
+            testlist = glob.glob(param)
+            if testlist:
                 # this is list of tests in local file
-                self.tests.append(param)
+                self.tests += testlist
```
phracek: Any reason, why not append?

jscotka: because glob.glob(...) creates list, ``param`` itselfs is string , thats why append has to be replaced by ``+=``

moduleframework/mtf_scheduler.py
```diff
@@ -219,31 +222,17 @@ def __init__(self, args, unknown):
...
-        avocadoAction = "avocado {ACTION} {AVOCADO_ARGS}".format(
-            ACTION=self.args.action, AVOCADO_ARGS=avocado_args)
-        bash = process.run("{AVOCADO} {a} {b}".format(
-            AVOCADO=avocadoAction, a=self.additionalAvocadoArg, b=" ".join(self.tests)), shell=True, ignore_status=True)
-        common.print_info(bash.stdout, bash.stderr)
-        common.print_debug("Command used: ", bash.command)
-        return bash.exit_status
+        avocadoAction = [self.AVOCADO, self.args.action] + avocado_default_args
+        command = subprocess.check_call(avocadoAction + self.additionalAvocadoArg + self.tests)
+        return command
```
phracek: This should be in exception like is mentioned here: https://docs.python.org/2/library/subprocess.html#subprocess.check_call

```
try:
  subprocess.check_call
except CalledProcessError as cpe
   return cpe.return_code
```

jscotka: you are right

Issue #84, opened 2017-09-06
[RFE] config.yaml should not be needed for testing


phracek:
Update meta-test-family, so config.yaml in some cases is not needed.

In case it is already implemented, document it in user_guide section, how developer / tester can use it.

jscotka:
should be fixed by
https://github.com/fedora-modularity/meta-test-family/pull/103
in case you not have your config.yaml, or configured via CONFIG envvar, there is used minimal config. it is good for base tests (or modules what does not provide any service)
but you have to set URL=to_some_link to not test default values in minimal config, byut what you want like:
```
MODULE=docker URL=docker.io/fedora avocado run your_test_or_modulelint
```

phracek:
@jscotka @dhodovsk I would like to test just dockerfile linter over Dockerfile. For this case URL is really not needed.

dhodovsk:
+1 for removing the need for URL from requirements of static linters. User might want to run linters on newly created/not pushed to registry Dockerfiles.

PR #200, opened 2018-01-18
Add support for static linters

phracek:

This Pull Request brings new argument into `mtf` command like --static-linters
which executes only helpmd_linter.py and dockerlint.py -> static-dockerlint-checks.py.


moduleframework/tools/static_dockerfile_checks.py
```diff
@@ -0,0 +1,85 @@
+from __future__ import print_function
+
+import os
+from avocado import Test
+from moduleframework import module_framework
+from moduleframework import dockerlinter
+
+class DockerInstructionsTests(module_framework.AvocadoTest):
```
jscotka: you've forgot  to change ``module_framework.AvocadoTest`` to ``Test``

moduleframework/tools/static_dockerfile_checks.py
```diff
@@ -0,0 +1,85 @@
...
+    dp = None
+
+    def setUp(self):
+        # it is not intended just for docker, but just docker packages are
+        # actually properly signed
+        self.dp = dockerlinter.DockerfileLinter()
+        if self.dp.dockerfile is None:
+            self.skip("Dockerfile was not found")
+
+    def tearDown(self, *args, **kwargs):
```
jscotka: when you change parent to ``Test`` you can remove this empty ``tearDown`` function

moduleframework/tools/static_dockerfile_checks.py
```diff
@@ -0,0 +1,85 @@
+from __future__ import print_function
+
+import os
+from avocado import Test
+from moduleframework import module_framework
```
jscotka: probably this import could be removed

moduleframework/common.py
```diff
@@ -91,6 +91,7 @@
 MODULE_DEFAULT_PROFILE = "default"
 TRUE_VALUES_DICT = ['yes', 'YES', 'yes', 'True', 'true', 'ok', 'OK']
 OPENSHIFT_INIT_WAIT = 50
+STATIC_CHECKS = ['static_dockerfiule_checks', 'helpmd_lint']
```
jscotka: wouldn't be better to add there some same pattern?
like you have: ``static_docker...`` and so that remane also helpmd_lint to same pattern.
otherwise there is typo ``...fiule...``

moduleframework/tools/static_dockerfile_checks.py
phracek: I use `assert_to_warn` function. AvocadoTest is needed. I don't see any reason, why not use AvocadoTest.

dhodovsk:
I tried this in container environment and everything runs as expected. There is still need to provide URL, but that issue is not blocking now and can be resolved in https://github.com/fedora-modularity/meta-test-family/issues/84. 

Thank you! :)

moduleframework/mtf_scheduler.py
```diff
@@ -194,9 +196,14 @@ class AvocadoStart(object):
...
     def __init__(self, args, unknown):
         # choose between TESTS and ADDITIONAL ENVIRONMENT from options
+        if args.static_linters:
+            self.tests += glob.glob("{MTF_TOOLS}/{STATIC_LINTERS}/*.py".format(
+                MTF_TOOLS=metadata_common.MetadataLoaderMTF.MTF_LINTER_PATH,
+                STATIC_LINTERS=common.STATIC_LINTERS))
         if args.linter:
-            self.tests += glob.glob("{MTF_TOOLS}/*.py".format(
-                MTF_TOOLS=metadata_common.MetadataLoaderMTF.MTF_LINTER_PATH))
+            self.tests += glob.glob("{MTF_TOOLS}/{GENERIC_TEST}/*.py".format(
```
jscotka: this change cause regression (now args.linter schedule all generic tests) with your change, just part of them will be sheduled. Is that intentional? we can discuss it. I'm not against this idea, but we should think about it more.

moduleframework/tests/static/dockerfile_lint.py
```diff
@@ -0,0 +1,82 @@
+from __future__ import print_function
+
+import os
+from avocado import Test
+from moduleframework import module_framework
```
jscotka: remove this line and remove ``module_framework.AvocadoTest`` from class parent

moduleframework/tests/static/dockerfile_lint.py
```diff
@@ -0,0 +1,82 @@
...
+from moduleframework import dockerlinter
+
+class DockerInstructionsTests(module_framework.AvocadoTest):
```
jscotka: replace by inherited from ``Test``

moduleframework/tests/static/dockerfile_lint.py
phracek: It is not possible, because we have in AvocadoTest function asser_to_warn.

PR #218, opened 2018-03-02
Mtf config file

jscotka:
This tries to remove hardcoded variables and move them to config file(s)


moduleframework/mtf.yaml
```diff
@@ -0,0 +1,61 @@
+default_module: docker
```
TomasTomecek: this file will require a lot of documentation so users would understand what each key means

jscotka: thanks for these comments.
Yep, I expect that package will provide some default configuration (for testing upstream/fedora)
and some users who understand that, is able to reconfigure it to another value. 

Yep, would be nice to add there documentation. for now probably it is better without documentation, than have these values hardcoded inside python code.

moduleframework/common.py
```diff
@@ -1,4 +1,4 @@
-#
+# -*- coding: utf-8 -*-
 # Meta test family (MTF) is a tool to test components of a modular Fedora:
```
phracek: I would prefer to remove it. It tests also containers. Not or modular Fedora. The line below too.

moduleframework/common.py
```diff
@@ -42,24 +42,82 @@
...
+    :return: None
+    """
+    if is_debug():
+        print_info(*args)
+
+
+
+class MTFConfParser(object):
+    configname = "mtf.yaml"
+    source_cfg_file = os.path.join(os.path.dirname(__file__), configname)
```
phracek: I would move it to `/usr/share/modularityframework/`. `/usr/lib/python2.7/... ` is for python packages.

moduleframework/common.py
```diff
@@ -42,24 +42,82 @@
...
+    default_cfg_file = os.path.join("/etc", configname)
```
phracek: Add `/etc` at the beginning of the file common.py. There is already set of constants.

moduleframework/common.py
```diff
@@ -42,24 +42,82 @@
...
+        config = {}
+        for cfgfile in [self.source_cfg_file, self.default_cfg_file, self.user_cfg_file]:
+            if os.path.exists(cfgfile):
+                print_info("MTF config loading: {}".format(cfgfile))
+                config.update(yaml.load(open(cfgfile)))
+        assert config.get("generic")
+        self.options = config
+
+
+    def get(self, section=None, key=None):
```
phracek: What about to use function `getattr` like https://docs.python.org/2/library/functions.html#getattr

moduleframework/common.py
```diff
@@ -868,9 +858,10 @@ def get_module_type():
     """
     amodule = os.environ.get('MODULE')
     readconfig = get_config(fail_without_url=False)
-    if "default_module" in readconfig and readconfig[
-        "default_module"] is not None and amodule is None:
+    if amodule is None and readconfig.get("default_module"):
```
phracek: what about something like:

`amodule = getattr(conf[options], "default_module", None)`

moduleframework/common.py
```diff
@@ -906,14 +897,14 @@ def get_docker_file(dir_name=DEFAULT_DIR_OF_DOCKER_RELATED_STUFF):
     if fromenv:
         dockerfile = fromenv
     else:
-        dockerfile = os.path.join(dir_name, DOCKERFILE)
+        dockerfile = os.path.join(dir_name, conf.get("docker","dockerfile"))
```
phracek: The same here.

moduleframework/common.py
```diff
@@ -906,14 +897,14 @@ def get_docker_file(dir_name=DEFAULT_DIR_OF_DOCKER_RELATED_STUFF):
...
     if not os.path.exists(dockerfile):
         print_debug("Dockerfile does not exists (you can use DOCKERFILE "
                     "envvar to set to another): %s" % dockerfile)
         dockerfile = None
     return dockerfile
 
-def get_helpmd_file(dir_name=DEFAULT_DIR_OF_DOCKER_RELATED_STUFF):
+def get_helpmd_file(dir_name=conf.get("docker","dockerfiledefaultlocation")):
     """
     Function returns full path to HelpMD file.
```
phracek: The same here.

moduleframework/common.py
```diff
@@ -868,9 +858,10 @@ def get_module_type():
     """
     amodule = os.environ.get('MODULE')
     readconfig = get_config(fail_without_url=False)
-    if "default_module" in readconfig and readconfig[
-        "default_module"] is not None and amodule is None:
+    if amodule is None and readconfig.get("default_module"):
         amodule = readconfig["default_module"]
+    if amodule is None and conf.options.get("default_module"):
+        amodule = conf.options["default_module"]
```
phracek: The same here.

moduleframework/helpers/openshift_helper.py
```diff
@@ -61,7 +62,7 @@ def __init__(self):
         random_str = ''.join(random.choice(string.lowercase) for _ in range(3))
         self.project_name = "{project}-{random_str}".format(project=self.app_name,
                                                             random_str=random_str)
-        common.print_debug(self.icontainer, self.app_name)
+        print_debug(self.icontainer, self.app_name)
```
phracek: Any idea, why you remove common? I thing, `common` means from which module `print_debug` is loaded.

moduleframework/helpers/openshift_helper.py
```diff
@@ -86,9 +87,9 @@ def _change_openshift_account(self, account="system:admin", password=None):
         :param password: Either user specified account
         """
         if password is None:
-            s = self.runHost("oc login -u %s" % account, verbose=common.is_not_silent())
+            s = self.runHost("oc login -u %s" % account, verbose=is_debug())
```
phracek: The same here with common.

moduleframework/helpers/openshift_helper.py
```diff
@@ -161,10 +162,10 @@ def _check_resource_in_json(self, json_output, resource='svc'):
                     # In metadata dictionary and name is stored pod_name
                     self.pod_id = json_output.get('metadata', {}).get('name')
                     return self.pod_id
-            elif resource == common.TEMPLATE:
+            elif resource == conf.get("openshift", "template"):
```
phracek: Either over `getattr` or over conf.get("openshift", None).get("template", None)

moduleframework/helpers/openshift_helper.py
```diff
@@ -315,7 +316,7 @@ def _verify_pod(self):
                  True, application is initiated and ready for testing
         """
         pod_initiated = False
-        for x in range(0, common.OPENSHIFT_INIT_WAIT):
+        for x in range(0, conf.get("openshift", "init_wait")):
```
phracek: twice `get` or even better over `getattr`.

moduleframework/helpers/rpm_helper.py
```diff
@@ -104,14 +107,16 @@ def setRepositoriesAndWhatToInstall(self, repos=[], whattooinstall=None):
         if repos:
             self.repos = repos
         else:
-            self.repos += get_compose_url() or self.get_url()
+            self.repos += common.get_compose_url() or self.get_url()
```
phracek: Here you use `common.`, but above only `print_info`. Any reason?
Keep the python style.

moduleframework/mtf.yaml
```diff
@@ -0,0 +1,104 @@
+# default module type in case you don't wath everytime use MODULE=docker or define it in config.yaml
+default_module: docker
+
+# openshift specific section
+openshift:
+  init_wait: 50
```
phracek: Add more documentation

moduleframework/mtf.yaml
```diff
@@ -0,0 +1,104 @@
...
+generic:
+# default architecture
+  arch: "x86_64"
+# command to find which packager to use (yum, dnf, apt-get)
+  packager_cmd: "test -e /usr/bin/dnf && echo 'dnf -y' ||
+   ( test -e /usr/bin/microdnf && echo 'microdnf'      ||
+   ( test -e /usr/bin/yum && echo 'yum -y'             ||
+   echo 'apt-get -y' ) )"
+# which values are considered as true
+  true_values: ['yes', 'YES', 'yes', 'True', 'true', 'ok', 'OK']
```
phracek: This can be one general function. I don't see any reason, why to have it here, in config file.

moduleframework/mtf_scheduler.py
```diff
@@ -205,10 +202,10 @@ def __init__(self, args, unknown):
         if args.linter:
             self.tests += glob.glob("{MTF_TOOLS}/{GENERIC_TEST}/*.py".format(
                 MTF_TOOLS=metadata_common.MetadataLoaderMTF.MTF_LINTER_PATH,
-                GENERIC_TEST=common.GENERIC_TEST))
+                GENERIC_TEST=conf.get("generic", "generic_tests")))
```
phracek: Again, either double `get` or even better `getattr`.

moduleframework/mtf.yaml
phracek: Configuration file is a bit long. What about to create a directory `/etc/mtf/` and have there several configuration files. One for each type. Ease for modification.

jscotka: this is good idea, but I'm not sure if anybody than me will modify it :-)
We can discuss it. I like this idea

moduleframework/common.py
jscotka: I thought about that too and in init version I had it there. I'll move it there

moduleframework/common.py
jscotka: I'll use that. sounds well

moduleframework/helpers/rpm_helper.py
jscotka: I'll fix it. would be nice also to fix all these issues in next PR, we can do issue for this cleanup

moduleframework/mtf.yaml
jscotka: I'm not able to add there documenttation, because you wrote this part and I've just moved these constants out of code to this file without knowledge what this does mean

moduleframework/mtf.yaml
jscotka: make sense

moduleframework/mtf.yaml
```diff
@@ -0,0 +1,103 @@
...
+  auth: ["odcs", "https://id.fedoraproject.org/openidc/", {Token: "Token", Authorization: "Authorization"},"odcs-authorizer","notsecret"]
+  scopes: ['openid', 'https://id.fedoraproject.org/scope/groups', 'https://pagure.io/odcs/new-compose', 'https://pagure.io/odcs/renew-compose', 'https://pagure.io/odcs/delete-compose']
+
+# ODCS section
+odcs:
+  compose_type: "module"
+  # timeout in secs, how long wait for compose (composes are cached, but first round can take long time (createrepo of packages))
+  timeout: 1200
+  new_compose_dict: { sigkeys: [''] }
+  url: "https://odcs.fedoraproject.org"
```
TomasTomecek: I expect that internally we'll have a different config.

jscotka: Yes, it is also mentioned few lines under (location of internal odcs server)
Probably we will ship, or somewhere stored config file for downstream testing

moduleframework/common.py
jscotka: fixed by elegant way:  config parser has dict as parent class, so all dict methods available there.

moduleframework/common.py
```diff
@@ -39,8 +38,35 @@
 import requests
 import warnings
 import ast
+from glob import glob
 from avocado.utils import process
-from moduleframework.mtfexceptions import ModuleFrameworkException, ConfigExc, CmdExc, DefaultConfigExc
+from mtfexceptions import ModuleFrameworkException, ConfigExc, CmdExc, DefaultConfigExc
+from core import DATADIR, is_debug, is_not_silent, print_info, print_debug
```
phracek: Why not only `import core`. print_info is so general and we don't know from which module comes from.

moduleframework/common.py
```diff
@@ -39,8 +38,35 @@
...
+
+
+
+class MTFConfParser(dict):
+    configdir = "mtf.conf.d"
+    source_cfg = os.path.join(*(["/"] + DATADIR + [configdir]))
```
phracek: Add `/` to DATADIR and remove ["/"].
I would write 
`os.path.join(*(DATADIR, [configdir]))`

What do you want to do with this code? All directories under DATADIR?

moduleframework/common.py
```diff
@@ -39,8 +38,35 @@
...
+    default_cfg = os.path.join("/etc", configdir)
+    user_cfg = os.path.expanduser("~/." + configdir)
+    pattern = "*.yaml"
+
+    def __init__(self):
+        config = {}
+        for cfgdir in [self.source_cfg, self.default_cfg, self.user_cfg]:
```
phracek: Will this work with several self.source_cfg directories?

moduleframework/common.py
```diff
@@ -252,29 +203,18 @@ def get_odcs_auth():
...
     """
-    odcstoken = os.environ.get('MTF_ODCS')
+    odcstoken = get_odcs_envvar()
 
     # in case you dont have token enabled, try to ask for openidc via web browser
-    if odcstoken in TRUE_VALUES_DICT:
+    if is_true(odcstoken):
+        if conf.get("openidc").get("token"):
+            # use value defined in config file if defined
+            return conf["openidc"]["token"]
```
phracek: `get` and `get` are missing.

moduleframework/common.py
```diff
@@ -252,29 +203,18 @@ def get_odcs_auth():
...
-        )
-
-        scopes = [
-            'openid',
-            'https://id.fedoraproject.org/scope/groups',
-            'https://pagure.io/odcs/new-compose',
-            'https://pagure.io/odcs/renew-compose',
-            'https://pagure.io/odcs/delete-compose',
-        ]
+        oidc = openidc_client.OpenIDCClient(*conf["openidc"]["auth"])
```
phracek: the same as above.

moduleframework/common.py
```diff
@@ -252,29 +203,18 @@ def get_odcs_auth():
...
+        scopes = conf["openidc"]["scopes"]
```
phracek: the same

moduleframework/common.py
```diff
@@ -348,7 +288,7 @@ def get_profile():
     :return: str
     """
 
-    return os.environ.get('PROFILE') or MODULE_DEFAULT_PROFILE
+    return os.environ.get('PROFILE') or conf["modularity"]["default_profile"]
```
phracek: the same.

moduleframework/common.py
```diff
@@ -469,7 +409,7 @@ def get_template(self):
         get location of template from config.yaml file
         :return: str
         """
-        return self.info.get(TEMPLATE)
+        return self.info.get(conf["openshift"]["template"])
```
phracek: Default value is missing.

moduleframework/helpers/openshift_helper.py
```diff
@@ -26,7 +26,8 @@
 import random
 import string
 from avocado.utils.process import CmdError
-from moduleframework import common
+from moduleframework.common import conf, is_debug, print_debug, print_info, trans_dict, sanitize_cmd, \
```
phracek: Some parts are part of module `core` instead of `common`.

moduleframework/helpers/openshift_helper.py
jscotka: because when using module common there, it hides errors when some bigger change, thats why I tries in more cases use these from common import ... things, what shows me, something what I've changed significantly (caused by this rewriting to use mtf config file)

moduleframework/common.py
jscotka: @phracek this is not possible, it is connected with using this in setup.py (there is just relative path, so I cannot add this to this DATADIR list)

moduleframework/common.py
jscotka: @phracek it is dict, so it is not important to use get, because it is already checked that is defined in `if`

moduleframework/common.py
```diff
@@ -24,23 +25,46 @@
...
-from moduleframework.mtfexceptions import ModuleFrameworkException, ConfigExc, CmdExc, DefaultConfigExc
+import glob
+import avocado.utils
+import mtfexceptions
+import core
+
+
+class MTFConfParser(dict):
+    configdir = "mtf.conf.d"
+    source_cfg = os.path.join(*(["/"] + core.DATADIR + [configdir]))
```
phracek: @jscotka What do you want to do with this command?

~~~
source_cfg = os.path.join(*(["/"] + ['etc']  + ["yum"]))
print source_cfg

prints /etc/yum
~~~

Why not only os.path.join(core.DATADIR, configdir)?
core.DATADIR could be 
~~~
core.DATADIR = "/usr/share/moduleframework"
~~~

moduleframework/common.py
```diff
@@ -252,36 +201,25 @@ def get_odcs_auth():
...
-        )
-
-        scopes = [
-            'openid',
-            'https://id.fedoraproject.org/scope/groups',
-            'https://pagure.io/odcs/new-compose',
-            'https://pagure.io/odcs/renew-compose',
-            'https://pagure.io/odcs/delete-compose',
-        ]
+        oidc = openidc_client.OpenIDCClient(*conf["openidc"]["auth"])
```
phracek: @jscotka Check if ["openidc"]["auth"] exists.
what about if oidc is None. Where do you catch traceback?

moduleframework/common.py
phracek: @jscotka You did not catch traceback. Is it enough exception for request.exception.HTTPError?

moduleframework/common.py
phracek: @jscotka This returns, either environment variable or value from `conf` or None. Is it requested?

moduleframework/common.py
phracek: @jscotka Do you count with None if `template` is not specified?

moduleframework/common.py
```diff
@@ -674,7 +612,7 @@ def run(self, command, **kwargs):
 
     def get_packager(self):
         if not self.packager:
-            self.packager = self.run(PACKAGER_COMMAND, verbose=False).stdout.strip()
+            self.packager = self.run(conf["generic"]["packager_cmd"], verbose=False).stdout.strip()
```
phracek: It will be specified always, right? What if packager_cmd is missing?

moduleframework/common.py
```diff
@@ -800,39 +738,39 @@ def get_config(fail_without_url=True, reload=False):
...
+                    raise mtfexceptions.Defaultmtfexceptions.ConfigExc("You have to use URL envvar for testing your images or repos")
                 cfgfile = "/usr/share/moduleframework/docs/example-config-minimal.yaml"
-                print_debug("Using default minimal config: %s" % cfgfile)
+                core.print_debug("Using default minimal config: %s" % cfgfile)
 
 
         try:
             with open(cfgfile, 'r') as ymlfile:
                 xcfg = yaml.load(ymlfile.read())
             doc_name = ['modularity-testing', 'meta-test-family', 'meta-test']
```
phracek: Why `meta-test`?

moduleframework/common.py
```diff
@@ -889,14 +828,14 @@ def get_module_type_base():
...
+            raise mtfexceptions.ModuleFrameworkException("Module (%s) does not provide parent backend parameter (there are: %s)" %
                                            (module_type, get_backend_list()))
     if parent not in get_backend_list():
-        raise ModuleFrameworkException("As parent is allowed just base type: %s" % get_backend_list)
+        raise mtfexceptions.ModuleFrameworkException("As parent is allowed just base type: %s" % get_backend_list)
     return parent
 
 
-def get_docker_file(dir_name=DEFAULT_DIR_OF_DOCKER_RELATED_STUFF):
+def get_docker_file(dir_name=conf["docker"]["dockerfiledefaultlocation"]):
```
phracek: This can be None if `dockerfiledefaultlocation` is empty. Do you count with it?

moduleframework/compose_info.py
```diff
@@ -40,7 +44,7 @@ class ComposeParser(object):
 
     def __init__(self, compose):
         self.compose = compose
-        xmlrepomd = compose + "/" + REPOMD
+        xmlrepomd = compose + "/" + common.conf["compose"]["repomd"]
```
phracek: Why not os.path.join?

moduleframework/core.py
```diff
@@ -0,0 +1,78 @@
...
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Authors: Jan Scotka <jscotka@redhat.com>
+#
+
+from __future__ import print_function
+import sys
+import os
+
+DATADIR = ['usr', 'share', 'moduleframework']
```
phracek: @jscotka Why not `/usr/share/moduleframework`?

moduleframework/common.py
```diff
@@ -24,23 +25,46 @@
...
 import ast
-from avocado.utils import process
-from moduleframework.mtfexceptions import ModuleFrameworkException, ConfigExc, CmdExc, DefaultConfigExc
+import glob
+import avocado.utils
+import mtfexceptions
+import core
+
+
+class MTFConfParser(dict):
```
phracek: @jscotka What about something like this:

~~~
class Config(dict):
     def __getattr__(self, key):
         try:
             return self[key]
         except KeyError:
             raise AttributeError
     def __setattr__(self, key, value):
         self[key] = value
     def __init__(self, yaml_file, release="current"):
         config = yaml.load(yaml_file)
         self["raw"] = config
         pdc = config["pdc"]
....
~~~

And you have access to all fields.

moduleframework/common.py
jscotka: @phracek I think that this is enough. Not important to catch it and transform to another exception.  And I've not changed this behaviour anyhow

moduleframework/common.py
jscotka: yes this is expected. or this also can cause traceback in case some user break this value from config. It is expected

moduleframework/common.py
jscotka: It has to be specified in this config. There is stored default value and in case user remove this item from config, it causes traceback, it is expected.

moduleframework/common.py
jscotka: because threre is by intention missing leading ``/`` so that it created relative path, instead of ``/usr/share/moduleframework``, and  also it is caused by that this syntax was used in ``setup.py``  for this directory. So I did not want to change it in setup.py, just put it out of hardcoded value in setup.py

moduleframework/common.py
jscotka: it causes traceback. by default it is defined in shipped mtf config. if user remove this. then it will lead to expected traceback.

moduleframework/common.py
jscotka: I don't know. I've defided it looong time ago, as shorter name in config files, it is closed to modularity-testing, it also miss word framework

moduleframework/common.py
jscotka: it will lead to traceback, it is expected

moduleframework/compose_info.py
jscotka: make sense. I'll replace it

moduleframework/core.py
jscotka: because original usage of this list is in ``setup.py`` in this format, copyied from other upstream projects, and I did't want to change this behaviour, just need to have this location in better place that in setup.py.

moduleframework/common.py
jscotka: @phracek  does that have  some benefits instead of actual implementation to redefine these getters and setters?
I trough to redefine ``__getattr__`` to lead some easier to debug exceptions (like, show path to file where is this found, and show value from default config, what should not be modifed), but it could be further improvement

phracek: yeah let's leave it for now.

PR #219, opened 2018-03-05
tag tests and add links to best practices

dhodovsk:
- Moved `fedora` and `rhel` tags from `DockerLabelsTests` to its test methods, to specify which label relates to which distro. 
 - Added links with test subject explanations

This is just quickly glued patch and the code should be improved in the future e.g by:
 - Refer to guidelines related to tested distro - when the tag is rhel do not refer to fedoraproject.org, but to the right guidelines in rhel or whatever relevant
 - Store links in separate place and add healthchecks to CI.

### DIFF
--- a/moduleframework/tests/static/dockerfile_lint.py
+++ b/moduleframework/tests/static/dockerfile_lint.py
@@ -22,16 +22,16 @@ class DockerInstructionsTests(module_framework.AvocadoTest):
             self.cancel("Dockerfile was not found")
 
     def test_from_is_first_directive(self):
-        self.assertTrue(self.dp.check_from_is_first(), msg="FROM instruction is not first.")
+        self.assertTrue(self.dp.check_from_is_first(), msg="FROM instruction is not first.\n\tMore info: https://docs.docker.com/engine/reference/builder/#from")
 
     def test_from_directive_is_valid(self):
-        self.assertTrue(self.dp.check_from_directive_is_valid(), msg="FROM instruction is not valid.")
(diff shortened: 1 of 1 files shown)


moduleframework/tests/static/dockerfile_lint.py
```diff
@@ -57,34 +57,75 @@ def _test_for_env_and_label(self, docker_env, docker_label, env=True):
...
+        :avocado: tags=fedora
+        """
+        self.assertTrue(self.dp.get_specific_label("maintainer"),
+                        msg=self._get_msg("Label maintainer",
+                                          reference='https://fedoraproject.org/wiki/Container:Guidelines#LABELS'))
 
     def test_release_label_exists(self):
-        self.assertTrue(self.dp.get_specific_label("release"), msg=self._get_msg("Label release"))
+        """
+        :avocado: tags=rhel,fedora
```
phracek: rhel is not needed. It is automatically generated by OSBS. I would prefer to remove it completely.

moduleframework/tests/static/dockerfile_lint.py
```diff
@@ -57,34 +57,75 @@ def _test_for_env_and_label(self, docker_env, docker_label, env=True):
             label_found = self.dp.get_specific_label(docker_label)
         return label_found
 
-    def _get_msg(self, label):
-        return label + " is missing in Dockerfile."
+    def _get_msg(self, label, reference=None):
+        msg = label + " is missing in Dockerfile."
+        if reference:
+            msg = msg + '\n\tMore info: ' +reference
```
phracek: Just space between `+` and reference.

phracek:
Just small fixes. Otherwise LGTM.

PR #224, opened 2018-03-09
before run avocado, list tests and find if there are some bad tests

jscotka:
example output:
```
mtf -l --url=fedora a c vs ds sd  /root
MTF config dir exists, search for /usr/share/moduleframework/mtf.conf.d/*.yaml
MTF config load: /usr/share/moduleframework/mtf.conf.d/mtf.yaml

ERROR: There are bad tests:
-------------
a: File not found ('a'; '/usr/share/avocado/tests/a')
c: File not found ('c'; '/usr/share/avocado/tests/c')
vs: File not found ('vs'; '/usr/share/avocado/tests/vs')
ds: File not found ('ds'; '/usr/share/avocado/tests/ds')
sd: Is a broken symlink
/root: Is not readable
```
ls output:
```
ls  a c vs ds sd  /root
ls: cannot access 'a': No such file or directory
ls: cannot access 'c': No such file or directory
ls: cannot access 'vs': No such file or directory
ls: cannot access 'ds': No such file or directory
sd

ls: cannot open directory '/root': Permission denied
```

### DIFF
--- a/moduleframework/mtf_scheduler.py
+++ b/moduleframework/mtf_scheduler.py
@@ -258,18 +258,16 @@ class AvocadoStart(object):
         # remove header line and remove last lines with is test types summary
         testlines = []
         for line in output.split("\n"):
-            if line.startswith(prefix_line) or not line.strip():
+            testline = line.strip()
+            if testline.startswith(prefix_line) or not testline:
                 continue
-            elif line.startswith(summary_line):
+            elif testline.startswith(summary_line):
                 break
             else:
-                testlines.append(line.strip())
-        for testline in testlines:
-            splitted = testline.split(" ", 1)
-            if splitted[0] in badstates:
-                badtests.append(splitted[1].strip())
+                splitted = testline.split(" ", 1)
+                if splitted[0] in badstates:
+                    badtests.append(splitted[1].strip())
         if badtests:
-
             core.print_info("", "ERROR: There are bad tests:", "-------------")
             core.print_info(*badtests)
             exit(19)
--- a/moduleframework/mtf_scheduler.py
+++ b/moduleframework/mtf_scheduler.py
@@ -197,7 +197,15 @@ class AvocadoStart(object):
     json_tmppath = None
     additionalAvocadoArg = []
     AVOCADO = "avocado"
-
+    A_KNOWN_PARAMS_SIMPLE=["-s", "-z", "-v", "-V",
+                           "--silent",
+                           "--show-job-log",
+                           "--replay-resume",
+                           "--filter-by-tags-include-empty",
+                           "--archive",
+                           "--journal",
+                           "--"
+                           ]
     def __init__(self, args, unknown):
         # choose between TESTS and ADDITIONAL ENVIRONMENT from options
         if args.linter:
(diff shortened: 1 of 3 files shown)


moduleframework/mtf_scheduler.py
```diff
@@ -209,16 +217,27 @@ def __init__(self, args, unknown):
...
         self.args = args
 
-        for param in unknown:
-            # take care of this, see tags for safe/unsafe:
-            # http://avocado-framework.readthedocs.io/en/52.0/WritingTests.html#categorizing-tests
-            testlist = glob.glob(param)
-            if testlist:
-                # this is list of tests in local file
+        # parse unknow options and try to find what parameter is test
+        while unknown:
```
phracek: Why not used argparse?

moduleframework/mtf_scheduler.py
```diff
@@ -229,21 +248,40 @@ def __init__(self, args, unknown):
         core.print_debug("additionalAvocadoArg = {0}".format(
             self.additionalAvocadoArg))
 
+    def check_tests(self):
+        output = subprocess.check_output([self.AVOCADO, "list", "-V", "--"] + self.tests)
+        assert "TEST TYPES SUMMARY" in output
+        badstates = ["NOT_A_TEST", "MISSING", "ACCESS_DENIED", "BROKEN_SYMLINK"]
+        badtests = []
+        # remove fist line, it is header, and remove last 11 lines, what is test summary
+        for testline in output.split("\n")[1:-11]:
```
phracek: avocado does not have something like `--no-header`. This looks like to me as RFE to avocado. What are last 11 lines?

moduleframework/mtf_scheduler.py
```diff
@@ -229,21 +248,40 @@ def __init__(self, args, unknown):
...
+        """
+
+        :param action: what avocado action to run: run, list, ...
+        :param avocado_default_args: list avocado additional argument
+        :return: return code of executed avocado command
+        """
         rc=0
         try:
-            subprocess.check_call(avocadoAction + self.additionalAvocadoArg + self.tests)
+            subprocess.check_call([self.AVOCADO, action] + avocado_default_args + self.additionalAvocadoArg + self.tests)
```
phracek: Can you use os.path.join? I would rewrite it to more pythonic style.

moduleframework/mtf_scheduler.py
jscotka: this behaviour was not change and it have sense to pass all unknown params to  avocado in case not known, like (--show-job-log) then it is possible to use mtf --show-job-log instread of create own param with same name and pass them to avocado.
create full argparser for each avocado param does not make sense. it is too complicated and avocado has plugin structure

moduleframework/mtf_scheduler.py
jscotka: this is avocado output (11 lines are this summary, it is mentioned  in comment "and remove last 11 lines, what is test summary")
```
$ avocado list -V a b c
Type       Test                                                  Tag(s)
NOT_A_TEST a: File not found ('a'; '/usr/share/avocado/tests/a') 
NOT_A_TEST b: File not found ('b'; '/usr/share/avocado/tests/b') 
NOT_A_TEST c: File not found ('c'; '/usr/share/avocado/tests/c') 

TEST TYPES SUMMARY
==================
ACCESS_DENIED: 0
BROKEN_SYMLINK: 0
EXTERNAL: 0
INSTRUMENTED: 0
MISSING: 0
NOT_A_TEST: 3
PYUNITTEST: 0
SIMPLE: 0
```

moduleframework/mtf_scheduler.py
jscotka: @phracek where to use ``os.path.join`` ?
I can rewrite it to some more python way, but I have not idea what is better?
``check_call`` uses ``dict`` as command, so I'm creating dict there with all command params

moduleframework/mtf_scheduler.py
phracek: And what if output is changed? We have to think about it for long time. I would prefer to discuss it with avocado team to provide computer readable format. E.g. JSON or YAML.

moduleframework/mtf_scheduler.py
```diff
@@ -229,21 +248,50 @@ def __init__(self, args, unknown):
...
+        # remove header line and remove last lines with is test types summary
+        testlines = []
+        for line in output.split("\n"):
+            if line.startswith(prefix_line) or not line.strip():
+                continue
+            elif line.startswith(summary_line):
+                break
+            else:
+                testlines.append(line.strip())
+        for testline in testlines:
```
phracek: This can be done inside `for line in output.split("\n")` cycle. New cycle is not needed.

jscotka: makes sense, thanks.

moduleframework/mtf_scheduler.py
```diff
@@ -209,16 +217,27 @@ def __init__(self, args, unknown):
...
 
-        for param in unknown:
-            # take care of this, see tags for safe/unsafe:
-            # http://avocado-framework.readthedocs.io/en/52.0/WritingTests.html#categorizing-tests
-            testlist = glob.glob(param)
-            if testlist:
-                # this is list of tests in local file
+        # parse unknow options and try to find what parameter is test
+        while unknown:
+            if unknown[0] in self.A_KNOWN_PARAMS_SIMPLE:
```
phracek: What is `unknown [0]`. I miss a documentation which will explain it.

phracek: This is '-' in self.A_KNOWN_PARAMS_SIMPLE which is always true.

jscotka: I didn't changed this, ``unknow`` variable were in previous version. I've just little bit chaged logic how to split arguments of avocado and what are tests.

PR #225, opened 2018-03-12
Fixes #220: If COPY or ADD not present, then True

### LINKED ISSUE #220: add/copy checks fail if the instructions are not present in a dockerfile
```
TEST FAIL:  /usr/lib/python2.7/site-packages/moduleframework/tests/static/dockerfile_lint.py:DockerInstructionsTests.test_copy_files_exist
     reason -> Some files in the COPY or ADD instruction do not exist.
     /home/zdravomil/avocado/job-results/job-2018-03-07T05.35-34c11fb/test-results/05-_usr_lib_python2.7_site-packages_moduleframework_tests_static_dockerfile_lint.py:DockerInstructionsTests.test_copy_files_exist/debug.log
```

The check fails even though there are no such instructions present in the dockerfile.

(I omitted internal info)

```
FROM rhel7:...

LABEL description="Red Hat OpenStack Platform..."
LABEL batch="..."
LABEL version-release="..."
LABEL io.openshift.tags="rhosp osp openstack..."
LABEL summary="Red Hat OpenStack Platform..."
LABEL io.k8s.display-name="Red Hat OpenStack Platform..."
LABEL com.redhat.component="openstack-kuryr-controller-docker"
LABEL name="rhosp.../openstack-kuryr-controller"
LABEL version="..."
LABEL architecture="x86_64"

RUN yum install -y ...

USER ...
ENTRYPOINT [ "/usr/bin/..." ]
```

phracek:

This issue fixes problem if COPY or ADD does not exists in Dockerfile.
For this case return has to be PASS of course.

### DIFF
--- a/moduleframework/dockerlinter.py
+++ b/moduleframework/dockerlinter.py
@@ -286,12 +286,15 @@ class DockerfileLinter(object):
         dir_name = os.getcwd()
         files = self._get_copy_add_files(os.path.dirname(self.dockerfile))
         f_exists = False
-        for f in files:
-            if f.startswith('http'):
-                f_exists = True
-                continue
-            if os.path.exists(os.path.join(dir_name, f)):
-                f_exists = True
-            else:
-                core.print_info("The file %s does not exist." % f)
+        if not files:
+            f_exists = True
+        else:
+            for f in files:
+                if f.startswith('http'):
+                    f_exists = True
+                    continue
+                if os.path.exists(os.path.join(dir_name, f)):
+                    f_exists = True
+                else:
+                    core.print_info("The file %s does not exist." % f)
         return f_exists
--- a/moduleframework/dockerlinter.py
+++ b/moduleframework/dockerlinter.py
@@ -286,6 +286,7 @@ class DockerfileLinter(object):
         dir_name = os.getcwd()
         files = self._get_copy_add_files(os.path.dirname(self.dockerfile))
         f_exists = False
+        missing_files = False
         if not files:
             f_exists = True
         else:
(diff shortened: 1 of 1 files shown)


moduleframework/dockerlinter.py
```diff
@@ -286,6 +286,9 @@ def check_copy_files_exist(self):
         dir_name = os.getcwd()
         files = self._get_copy_add_files(os.path.dirname(self.dockerfile))
         f_exists = False
+        if not files:
+            f_exists = True
+            return f_exists
```
jscotka: I suggest:
to use there ``return True`` or maybe better will be to have there something like:
```
if not files:
    f_exists = True
else:
    for f in files:
        if f.startswith('http'):
    ...
return  f_exists
```

moduleframework/dockerlinter.py
```diff
@@ -286,12 +286,15 @@ def check_copy_files_exist(self):
...
+            f_exists = True
+        else:
+            for f in files:
+                if f.startswith('http'):
+                    f_exists = True
+                    continue
+                if os.path.exists(os.path.join(dir_name, f)):
+                    f_exists = True
+                else:
+                    core.print_info("The file %s does not exist." % f)
```
TomasTomecek: I would expect `f_exists = False` here and would not even override it: if any file is missing then the check should fail.

moduleframework/dockerlinter.py
```diff
@@ -286,12 +286,20 @@ def check_copy_files_exist(self):
...
-            if f.startswith('http'):
-                f_exists = True
-                continue
-            if os.path.exists(os.path.join(dir_name, f)):
-                f_exists = True
-            else:
-                core.print_info("The file %s does not exist." % f)
-        return f_exists
+        missing_files = False
+        if not files:
```
TomasTomecek: to me, this code block is still way too complicated; but it's good enough from my PoV

PR #230, opened 2018-03-14
Fix issue with VOLUME and add VOLUME dockerlint tests

### LINKED ISSUE #222: Linters: all fails when volume value not quoted
All linter tests fail with `ERROR:  invalid syntax (<unknown>, line 1)` on images that have unquoted `VOLUME` value.
Test data:
```
FROM awesome/fedora

LABEL com.redhat.component="comp" \
      name="trulo" \
      version="7.0.0" \
      architecture="x86_64" \
      summary="Just keep swimming"

VOLUME /opt/trulo/data
```

phracek:

This issue fixes #222  and brings more tests for linters.

### DIFF
--- a/moduleframework/dockerlinter.py
+++ b/moduleframework/dockerlinter.py
@@ -33,10 +33,10 @@ class DockerfileLinter(object):
 
     dockerfile = None
     oc_template = None
-    dfp_structure = {}
-    docker_dict = {}
 
     def __init__(self):
+        self.dfp_structure = {}
+        self.docker_dict = {}
         dockerfile = common.get_docker_file()
         if dockerfile:
             self.dockerfile = dockerfile
@@ -60,7 +60,12 @@ class DockerfileLinter(object):
 
     def _get_volume(self, value):
         """Function evaluates a value and returns as string."""
-        return get_string(value)
+        try:
+            value = ast.literal_eval(value)
+        except SyntaxError:
+            # Convert string to list
+            return [value]
+        return value
 
     def _get_label(self, val):
         """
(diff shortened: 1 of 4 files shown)


moduleframework/dockerlinter.py
```diff
@@ -33,10 +33,10 @@ class DockerfileLinter(object):
 
     dockerfile = None
     oc_template = None
-    dfp_structure = {}
-    docker_dict = {}
 
     def __init__(self):
+        self.dfp_structure = {}
+        self.docker_dict = {}
```
jscotka: why you moved this from previous lines here?
Is this more pythonish? I don't know, just want to know answere :-)

moduleframework/dockerlinter.py
```diff
@@ -60,7 +60,12 @@ def _get_env(self, value):
 
     def _get_volume(self, value):
         """Function evaluates a value and returns as string."""
-        return get_string(value)
+        try:
+            value = ast.literal_eval(value)
+        except SyntaxError:
+            # Convert string to list
+            return [value]
```
jscotka: is this logic okay?
I don't know what is proper syntax, but I can imagine that there is possible:
```
In [4]:  ast.literal_eval("'abc'")
Out[4]: 'abc'

In [5]:  ast.literal_eval("['abc']")
Out[5]: ['abc']

In [6]:  ast.literal_eval("abc")
ValueError                                Traceback (most recent call last)
...
```
just last one caused error, but in case there will be 'abc' it will not be any error and it returns string.
Is this expected? syntax like ``VOLUME "/etc"`` or ``VOLUME '/etc'`` is not valid?

moduleframework/dockerlinter.py
phracek: There was initialization problem. `docker_dict` was not empty for some reason. Nothing else.

moduleframework/dockerlinter.py
phracek: Good point with ValueError. I will add it.

